PR #71, opened 2026-07-04
Add LinkedIn job search provider

amikai:
## Summary
- Adds `internal/provider/linkedin/`, a hand-written HTML client + parser for LinkedIn's guest job search (`jobs-guest/.../seeMoreJobPostings/search`) and job detail (`jobs/view/{id}`) endpoints — same family as `google`/`tsmc`/`synopsys` (SSR HTML, no `ogen` codegen) since neither LinkedIn endpoint returns JSON.
- `openapi.yaml` documents both endpoints, with every observed enum value spelled out (`f_WT` workplace type, `f_JT` job type), and registers in the Makefile's `OPENAPI_SPECS` validation list.
- `testdata/` holds real captured fixtures (one search response, one detail response) plus the exact `curl` commands used to get them; all unit tests run against a `mocksrv` serving those fixtures, never live LinkedIn.
- Adds `cmd/linkedin/main.go`, a CLI mirroring the other providers' pattern.

## Notable implementation details
- A cold request to `jobs/view/{id}` returned HTTP **999** (LinkedIn's bot-suspected/authwall response) during fixture capture; it succeeded once cookies from the prior search request were carried along. `NewClient` now defaults to a cookie-jar-backed `http.Client` for this reason, with a regression test (`TestGetHTMLBotBlocked`) and a note in `openapi.yaml` (999 isn't a valid OpenAPI response key, so it's documented under `default` per spec).
- `cmd/linkedin/main.go`'s `--fetch-details` defaults to `0` (off): the detail endpoint is the more block-prone of the two, so fetching full descriptions for every search result is opt-in rather than automatic.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (all packages pass, including 8 new tests in `internal/provider/linkedin`)
- [x] `make validate-openapi` (all specs, including the new one, pass `pythonopenapi/openapi-spec-validator`)
- [x] `gofmt -l` clean on all new files


cmd/linkedin/main.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func writeReport(w io.Writer, keywords string, search *linkedin.JobsResponse, details map[string]*linkedin.JobDetailResponse) {
+	fmt.Fprintf(w, "LinkedIn Jobs Report\n")
+	fmt.Fprintf(w, "Keywords: %s\n", keywords)
+	fmt.Fprintf(w, "Found %d jobs\n\n", len(search.Jobs))
+
+	for i, job := range search.Jobs {
+		fmt.Fprintf(w, "%d. [%s] %s\n", i+1, job.ID, job.Title)
+		fmt.Fprintf(w, "URL: https://www.linkedin.com/jobs/view/%s\n", job.ID)
```
amikai: `writeReport` hardcodes `https://www.linkedin.com/jobs/view/%s` here instead of using the `--base-url` flag value. The flag is threaded correctly into the actual HTTP client (`linkedin.NewClient(*baseURL, nil)` at line 54), but `writeReport`'s signature never receives `baseURL`, so a user who overrides `--base-url` (e.g. to point at `linkedin.NewMockServer()` for local testing) gets a report whose printed "URL:" lines still say `linkedin.com` regardless of where the data actually came from.

Sibling CLIs already do this correctly — `cmd/tsmc/main.go:61` and `cmd/google/main.go:136` both interpolate the live `baseURL` value into their printed URLs. Consider passing `baseURL` into `writeReport` here too.

amikai: Fixed in 02c1d2d — `writeReport` now takes `baseURL` and interpolates it the same way `cmd/google`/`cmd/tsmc` do.

amikai:
## Code Review

The PR adds a hand-written HTML client + parser for LinkedIn's guest job-search and job-detail endpoints, following the same SSR-scraping pattern as the google/tsmc/synopsys providers, plus a CLI, OpenAPI doc, and mocksrv-based tests against real captured fixtures. The overall structure is solid and consistent with the sibling providers; the tests genuinely exercise the captured HTML. Eight findings survived adversarial verification (42 raw candidates from 8 review angles, deduplicated and each verified against the actual code and fixtures) — two confirmed correctness bugs, one design gap that will bite when the provider is wired into the MCP server, three plausible parser-robustness issues, and two confirmed duplication cleanups.

### Correctness / design

1. `internal/provider/linkedin/client.go:160` — A cold `JobDetail` call still hits the HTTP 999 authwall; the cookie jar only helps if `Jobs()` ran first on the same client. **CONFIRMED.** It works today only because `cmd/linkedin` always searches before fetching details. Every sibling provider is registered in the MCP server as a standalone `<provider>_get_job_detail` tool, so the moment linkedin gets the same wiring, cold detail calls will fail with 999 on most invocations. Consider a warm-up request (or one retry-after-priming) inside `JobDetail`/`getHTML` itself.

2. `internal/provider/linkedin/parse.go:171` — `Remote` on job detail is a substring scan of the entire description, so any mention of "remote" flips it, and search vs detail disagree for the same job. **CONFIRMED.** `looksRemote` does plain `strings.Contains` for `remote`/`wfh`; a description saying "this is not a remote position" yields `Remote=true` on detail while the same job's search card (which only scans title+location) says `false`. Either scope detail to title+location too, or rename the field to reflect that it's a hint.

3. `cmd/linkedin/main.go:90` — `--company-ids` is split on `,` without trimming, silently breaking the filter. **CONFIRMED.** `--company-ids "1441, 162479"` produces `f_C=1441,+162479` (the space gets percent-encoded by `q.Encode()`), so the second company silently matches nothing. One `strings.TrimSpace` per part fixes it.

4. `internal/provider/linkedin/parse.go:102` — `jobIDFromHref` takes everything after the last hyphen with no numeric validation, ignoring the `data-entity-urn="urn:li:jobPosting:<id>"` attribute present on all 10 fixture cards. **PLAUSIBLE.** A trailing-slash href yields ID `4422697744/`; a slug-less canonical href (`/jobs/view/4422697744`) yields `""` and the job is silently dropped. The urn attribute is a strictly more robust source that's already on the very div the parser matches.

5. `internal/provider/linkedin/client.go:196` — A block page served 200 at any path other than `signup`/`authwall` (e.g. `/checkpoint`) makes `Jobs()` return an empty result with nil error. **PLAUSIBLE.** `parseJobsHTML` finding zero `base-search-card` divs is indistinguishable from a genuinely empty search — note `JobDetail` does error ("not found in response") in the analogous case, so search and detail have asymmetric block-detection.

6. `internal/provider/linkedin/parse.go:192` — `parseCriteria` takes the first `<span>` anywhere in the criteria `<li>`, ignoring the `description__job-criteria-text--criteria` class that this PR's own openapi.yaml documents as the selector. **PLAUSIBLE.** Any decorative/sr-only span LinkedIn inserts before the value span would silently corrupt `SeniorityLevel`/`EmploymentType`/`Industries`. Matching the documented class (with bare-span fallback) aligns the parser with its own spec.

### Duplication (both CONFIRMED byte-identical)

7. `internal/provider/linkedin/parse.go:272` — `hasClass`/`attrVal`/`textContent`/`appendNodeText` are the 4th copy of the same helpers (md5-identical to google's; tsmc and synopsys carry 3 of 4 each). Drift is already visible next door: linkedin's client pins `Chrome/120` as User-Agent while google/tsmc/synopsys all pin `Chrome/124`. At four consumers, an `internal/htmlutil` package is due.

8. `cmd/linkedin/main.go:157` — `labels()`/`usageWithChoices()` (including the subtle `""`-sentinel workaround for `ffval.Enum`) are roughly the 5th copy across cmd/ mains, and a downgrade from the generic `map[string]V` versions in cmd/tsmc and cmd/nvidia that already handle `map[string]string`. Worth hoisting into a shared helper.

---
Verified but dropped: the shared `--timeout` context across search + N detail fetches is real but exactly mirrors cmd/google and cmd/tsmc, so it's repo convention rather than a regression; the greedy `\?url=([^"]+)` apply-URL regex and the first-`h3` title fallback are real mechanisms but no captured fixture exhibits the triggering markup.

PR #73, opened 2026-07-05
Add generic Workday CXS provider and CLI (cmd/workday)

amikai:
## Summary

Adds generic Workday CXS ("myworkdayjobs.com") support in two layers, both
tenant-agnostic (any company, no hardcoded facet ids or server URL):

1. **`internal/provider/workday/`** — a reusable `openapi.yaml` describing the
   Workday Candidate Experience (CXS) API contract shared across tenants, plus
   the generated ogen client (same layout as `nvidia`/`google`/`tsmc`/`cake`).
2. **`cmd/workday/`** — a CLI that searches and inspects job postings on any
   Workday CXS career site given only its base URL. It's the generic
   counterpart to `cmd/nvidia`, which is hardcoded to one tenant's facet
   vocabulary and server URL.

## Why

`internal/provider/nvidia` bakes in NVIDIA-specific facet GUIDs and a fixed
server URL. This PR factors out the parts of the Workday CXS contract that
hold across tenants and adds a CLI to exercise them, as groundwork for
onboarding additional Workday-based companies without writing per-tenant code.

## API contract (provider layer)

- **Multi-tenant URL shape:** `{tenant}.{wdInstance}.myworkdayjobs.com/wday/cxs/{tenant}/{site}`.
  Because the base URL is entirely tenant-specific, the generated client takes
  `serverURL` as a runtime parameter — nothing is hardcoded.
- **Endpoints:** `POST /jobs` (search/filter/paginate) and
  `GET /job/{location}/{titleSlug}` (detail). `externalPath` from a search
  result must be split into two path segments — a single combined path
  parameter fails because URI encoders percent-encode the `/`.
- **Facets are tenant-specific and self-describing.** `appliedFacets` filters
  by facet-parameter-name → array of opaque Workday GUIDs, all configured per
  tenant. **Confirmed live against NVIDIA:** a single unfiltered `POST /jobs`
  returns the tenant's *complete* current facet tree in one call — 4 top-level
  groups including a nested `locationMainGroup` (location type → 34 countries →
  121 city-level sites), each with a resolvable id and live count. Modeled as a
  recursive `FacetNode` (group vs. leaf). This is why the CLI can discover a
  new tenant's entire filter vocabulary with one request.
- Caveats note what's confirmed platform-wide vs. only verified against
  NVIDIA's tenant (the 2000 pagination cap, error codes, the split-path
  requirement).

## CLI (`cmd/workday`)

Built on `ff/v4`'s `Command` tree; `--base-url`/`--timeout`/`--format` are
shared root flags, with two subcommands:

```
workday --base-url URL facets [--search-text S] [--facet name=id ...] [--format text|json]
workday --base-url URL search [--search-text S] [--limit N] [--offset N] [--facet name=id ...] [--format text|json]
```

- **`facets`** — discovers a tenant's current facet tree via a `Limit: 1`
  unfiltered call and prints it (indented text tree, or JSON). This is how a
  user learns the opaque facet ids to filter by.
- **`search`** — searches, then fetches full detail for each result (mirrors
  `cmd/nvidia`'s report behavior). A per-job detail-fetch failure is
  non-fatal: it falls back to a public site URL derived via the new
  `PublicSiteURL` helper and records the error instead of a description.
- `--facet name=id` is repeatable — same name OR's within a facet, different
  names AND across facets, matching `AppliedFacets`'s `map[string][]string`
  shape 1:1.

**Provider additions to support the CLI:** `path.go` with `SplitExternalPath`
(ported from `nvidia`) and `PublicSiteURL` (derives the public career-site
origin from the CXS base URL's last path segment; confirmed against NVIDIA's
`nvidiaSiteURL` constant).

## Not included

- Not wired into `internal/openingsmcp` (no MCP tool layer).
- Does not migrate `cmd/nvidia` / `internal/provider/nvidia` to the generic
  package.
- No auto-pagination — `search` fetches one page per invocation, like `cmd/nvidia`.

## Test plan

- [x] `go generate ./internal/provider/workday/...` runs clean
- [x] `go test ./...` passes (new `path_test.go` + `client_test.go` with synthetic fixtures)
- [x] `go build ./... && go vet ./...` clean (no regressions elsewhere)
- [x] Live smoke tests against NVIDIA's tenant: `facets` (text + JSON) renders
      the full nested tree; `search` returns detail-enriched results; `--facet`
      filtering, missing `--base-url`, and malformed `--facet` all behave correctly


amikai:
## Self-review (multi-angle pass over the hand-written surface)

Generated ogen code excluded from scope. Findings most-severe first, with what I'm doing about each.

### Fixing

**1. `internal/provider/workday/openapi.yaml` — `JobsResponse` marks `facets` required.**
`required: [total, jobPostings, facets]` makes the generated decoder/validator reject any `/jobs`
response that omits `facets` (or sends `facets: null`), so **both** `search` and `facets` abort even
when `total`/`jobPostings` are present and fetchable. This diverges from the live-confirmed NVIDIA
reference, whose `JobsResponse` deliberately requires only `[total, jobPostings]`. → dropping `facets`
from `required` and regenerating.

**2. `cmd/workday/main.go:288-289` — detail values unconditionally overwrite good summary values.**
`r.PostedOn = info.PostedOn.Value` reads `.Value` without gating on `.Set`; `postedOn` is optional, so a
detail response that omits it clobbers the summary's `"Posted Today"` with `""` (then dropped from both
text and JSON output). `r.Title = info.Title` has the same failure mode for an empty title. → gate both:
overwrite only when the detail actually provides a value.

**3. `cmd/workday/main.go:217-220` — sparse postings (empty `externalPath`) silently dropped.**
The loop `continue`s on empty `externalPath`, so such postings vanish from the report and `showing N`
under-counts — contradicting the docstring's claim that it "mirrors cmd/nvidia's report behavior
exactly" (nvidia lists them as "no detail available"). → stop skipping; list them with a
"no detail available" note, matching nvidia.

**4. `cmd/workday/main.go:153` — `printFacetNode` uses `len(node.Values) > 0` to tell group from leaf.**
A group node with a present-but-empty `values: []` is misclassified as a leaf and prints a meaningless
`  id=  count=0` line. → discriminate on `node.FacetParameter.Set` (groups carry it, leaves don't).

**5. `internal/provider/workday/client_test.go` — no fixture exercises an omitted required field.**
Every fixture is a fully-populated happy path, so the `facets`-omitted breakage in #1 had zero regression
guard. → adding a test asserting the client tolerates a `/jobs` response with `facets` omitted.

### Considered, keeping as-is

**`openapi.yaml:331` — `JobPostingInfo` requires `[title, jobDescription]`.** A tenant omitting
`jobDescription` on the detail endpoint would fail to decode (caught per-job, degrades to
"job detail unavailable"). But the live-confirmed NVIDIA reference requires the same two fields, and the
transient-omission behaviour documented in this spec was observed on the *list* endpoint (`JobSummary`),
not detail. Relaxing it would diverge from the proven reference without evidence, so I'm leaving it.

### Noted, deferred (pure DRY, no behaviour change)

`runFacets`/`runSearch` share a byte-identical prologue and JSON-encode idiom, and the two error-fallback
tails in `fetchJobResult` repeat. Low priority; left out to keep this change focused on correctness.

PR #77, opened 2026-07-05
Add internal/provider/greenhouse: ogen client for the Job Board API

amikai:
## Summary
- Scaffolds a new Greenhouse provider (`internal/provider/greenhouse`) covering exactly the two public Job Board API endpoints: `GET /boards/{board_token}/jobs` (list) and `GET /boards/{board_token}/jobs/{job_id}` (detail), client generated via `ogen` from `openapi.yaml`.
- The spec was corrected against live API responses (not just the published docs): `JobDetail` was missing `departments`/`offices` entirely; `ai_disclaimer`, `include_ai_disclaimer`, `ai_opt_out_request_url`, `compliance`, and `demographic_questions` are nullable but weren't marked as such; an undocumented `education` enum field (`education_required`/`education_optional`) appears on some boards.
- `client_test.go` uses real fixtures captured from live boards (Safari AI's list, Anthropic's detail with/without `questions`/`pay_transparency`) with literal struct comparisons, plus 404 cases for unknown board/job.
- Added the new spec to the `validate-openapi` Makefile target.

## Out of scope
MCP tool wiring (`greenhouse_search_jobs` / `greenhouse_get_job_detail` in `cmd/openings-mcp`) is tracked as a follow-up task and not included in this PR — this PR is the provider client only.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean on all new files
- [x] `go test ./...` (all packages pass, including 5 new Greenhouse tests: list, list-404, detail, detail-with-questions/pay-transparency, detail-404)
- [x] `make validate-openapi` (Docker-based OpenAPI validator, all 8 specs including the new one)


amikai:
Found one blocking schema/runtime mismatch:

`internal/provider/greenhouse/openapi.yaml` models `location_questions`, `questions`, `metadata`, and `pay_input_ranges` as non-null arrays in `JobDetail`, but live Greenhouse job detail responses can return these fields as explicit `null`. I verified this against `GET https://boards-api.greenhouse.io/v1/boards/safariai/jobs/5019966007` on July 5, 2026 UTC, which returned:

```json
{
  "metadata": null,
  "location_questions": null,
  "questions": null,
  "compliance": null,
  "demographic_questions": null,
  "pay_input_ranges": null
}
```

`compliance` and `demographic_questions` are already nullable in the spec, but the other four generated decoder paths call `d.Arr(...)`, so valid Greenhouse responses can fail during decode before callers can use the provider. Please mark those fields nullable, regenerate the `ogen` client, and add a fixture/test that covers explicit nulls.

amikai:
## Review of PR #77 — Greenhouse provider

The PR adds a clean, well-tested ogen client for the Greenhouse Job Board API, and the live-capture methodology it describes is genuinely good. The problem is that the methodology wasn't applied to every spec path: verification against the live API (all findings below were checked independently, several reproduced with real requests) found three decode bugs that make the client fail outright on real boards, all in paths the fixtures don't cover.

### Confirmed correctness bugs, most severe first

1. **[openapi.yaml:112](internal/provider/greenhouse/openapi.yaml#L112)** — `Office.location` is a non-nullable string, but the live API returns `"location": null` for offices without a location. Reproduced live: `ListJobs(content=true)` on safariai fails with `decode field "location": unexpected byte 'n'` (2 of 7 offices are null). `GetJob` fails the same way on any board with a location-less office.

2. **[openapi.yaml:382](internal/provider/greenhouse/openapi.yaml#L382)** — `JobDetail.metadata` is a non-nullable array while `JobSummary.metadata` is correctly `nullable: true`. The detail endpoint also returns `"metadata": null` — reproduced live on safariai job 5019966007, whose `GetJob` fails with a decode error today.

3. **[openapi.yaml:393](internal/provider/greenhouse/openapi.yaml#L393)** — `JobDetail` metadata item `value` is typed string-only, but real boards return `null` (duolingo, unset fields) and arrays (cloudflare, multi_select) — both reproduced live; either fails the whole `GetJob` decode.

4. **[openapi.yaml:146](internal/provider/greenhouse/openapi.yaml#L146)** — Closed enums on response fields (`education` at 146 and 343, `QuestionField.type` at 199, `DemographicQuestion.type` at 303): the generated response decoder calls `Validate()`, which hard-fails the **entire** ListJobs/GetJob call on any value outside the hand-observed list. `education` is undocumented upstream, and the `DemographicQuestion` enum was never exercised (both fixtures had `demographic_questions: null`). No sibling spec puts enums on response fields — they're all request-side. Plain `type: string` with observed values in the description matches both the convention and the risk profile.

5. **[mocksrv.go:25](internal/provider/greenhouse/mocksrv.go#L25)** — The `content=true` list path is completely untested: the mock ignores the query string, no `content=true` fixture was captured, and no test sets `Content`. This is exactly where bug #1 hides — the one spec path with a confirmed live failure passes `go test` green. Capturing a `content=true` fixture would have caught #1.

6. **[mocksrv.go:28](internal/provider/greenhouse/mocksrv.go#L28)** — The detail mock dispatches only on `questions=true` and never reads `pay_transparency`. Since the only test setting `PayTransparency` also sets `Questions`, a regression that silently stops encoding `pay_transparency` on the wire would still pass the whole suite; the mock also serves the wrong fixture for any future pay_transparency-only request.

7. **[openapi.yaml:38](internal/provider/greenhouse/openapi.yaml#L38)** — Only 200/404 are declared and the 404 has no body schema, so `ListJobsNotFound`/`GetJobNotFound` are empty structs and the live 404 body (`{"status":404,"error":"Job not found"}`) is discarded; 429/5xx surface as bare `unexpected status code`. Both cake and job104 specs declare a `default` response with a typed `ErrorResponse` — worth matching before the MCP wiring needs to report errors.

8. **[testdata/jobs_req.sh:3](internal/provider/greenhouse/testdata/jobs_req.sh#L3)** — The capture scripts pipe through `jq .` (pretty-print) but the committed fixtures are single-line compact JSON, so the scripts as committed cannot have produced the committed files; re-running one flips the fixture to multi-line. Siblings use the same scripts but commit pretty-printed fixtures — committing pretty-printed output here would make scripts and fixtures agree.

### Minor cleanups (not blocking)

- The identical GDPR `DataCompliance` literal is copy-pasted 8× in `client_test.go` (hoist to one package-level var, as cake's tests do with `wantSearch`/`wantDetail`).
- The two explicit not-found mock handlers are redundant — `ServeMux` already 404s unregistered paths and the generated decoders never read the 404 body, so both handlers can simply be deleted.
- `serveMockJSON` is a fifth byte-identical copy across providers, but that matches the existing four-provider convention — a separate cross-provider refactor, not a blocker here.

### Checked and refuted (no action needed)

`pay_input_ranges` never comes back `null` (always `[]`, checked on 5 live boards), and the `ai_*` fields do not appear on the list endpoint (checked live) — the spec's summary/detail split there is correct.

The common thread for fixes 1–3: recapture fixtures from boards that exercise the gaps (a `content=true` list, a detail with `metadata: null`, and one with non-string metadata values), mark the three spec fields nullable/loosely-typed, regenerate, and the existing test style covers the rest.

PR #80, opened 2026-07-06
feat: add Lever Postings API provider

amikai:
## Summary

Adds `internal/provider/lever` — an ogen-generated client for the public, unauthenticated [Lever Postings API](https://github.com/lever/postings-api), following the same provider-layer pattern as Workday/LinkedIn (no MCP wiring, no CLI yet).

- **OpenAPI spec** covering the JSON consumer surface: `GET /v0/postings/{site}` (skip/limit + case-sensitive repeatable `location`/`commitment`/`team`/`department` filters, `level`) and `GET /v0/postings/{site}/{postingId}`, on both the global and EU instances. `mode` is pinned to a single-value `json` enum so responses can never fall back to HTML. The apply endpoint, `group` parameter, and HTML/iframe embedding features are deliberately out of scope.
- **Schema grounded in the live API**: `createdAt` (epoch ms, returned but undocumented upstream) added; everything except `id`/`text` optional since salary fields are omitted on postings without salary data. Error shape pinned to the observed 404 body `{ok:false, error:"Document not found"}`.
- **Tests**: mock server replaying real `leverdemo` fixtures (captured by checked-in curl scripts); tests pin request construction (`mode=json` always sent, repeated filters encoded as `location=A&location=B`), response decoding, and 404 decoding for unknown site/posting.
- **Curated tenant list**: `companies.yaml` with 16 companies (Palantir, Mistral AI, Binance, Zoox, Ninja Van, …), each verified on 2026-07-06 to return a non-empty postings array; embedded and exported directly from `companies.go`.
- Registered the spec in `make validate-openapi`.

Design spec and implementation plan are included under `docs/superpowers/`.

## Test plan

- [x] `go test -race -vet=all ./...` — all packages pass
- [x] `make validate-openapi` — all 9 specs OK
- [x] `go generate ./internal/provider/lever/...` produces no diff (generated code in sync with spec)


internal/provider/lever/client_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	cats := first.Categories.Value
+	assert.Equal(t, NewOptString("Arlington, TX"), cats.Location)
+	assert.Equal(t, NewOptString("Regular Full Time (Salary)"), cats.Commitment)
+	assert.Equal(t, NewOptString("Professional Services"), cats.Team)
+	assert.Equal(t, NewOptString("Customer Success"), cats.Department)
+	assert.Equal(t, []string{"Arlington, TX"}, cats.AllLocations)
+
+	require.NotEmpty(t, first.Lists)
+	assert.Equal(t, "Qualifications", first.Lists[0].Text)
```
amikai: make want as literal structure to compare it

internal/provider/lever/client_test.go
```diff
@@ -0,0 +1,134 @@
...
+	c, err := NewClient(srv.URL, WithClient(srv.Client()))
+	require.NoError(t, err)
+
+	got, err := c.GetPosting(t.Context(), GetPostingParams{
+		Site:      "leverdemo",
+		PostingId: "33538a2f-d27d-4a96-8f05-fa4b0e4d940e",
+	})
+	require.NoError(t, err)
+	assert.Equal(t, "33538a2f-d27d-4a96-8f05-fa4b0e4d940e", got.ID)
+	assert.Equal(t, "AbelsonTaylor Writer", got.Text)
```
amikai: make want as literal structure to compare it

internal/provider/lever/companies_test.go
amikai, line 1: Delete companies_test.go. don't need this

PR #84, opened 2026-07-06
feat: add cmd/lever CLI

amikai:
## Summary

Adds `cmd/lever`, the CLI companion to the Lever provider from #80 (stacked on that branch — retarget to `main` after #80 merges).

- Mirrors `cmd/workday`'s shape: single `main.go`, `ff/v4` command tree, root flags `--site`/`--timeout`/`--format text|json`.
- Subcommands: `companies` (lists the 16 curated sites, no network), `search` (repeatable case-sensitive `--location`/`--commitment`/`--team`/`--department` filters plus `--level`/`--limit`/`--skip`, one page per invocation), and `get POSTING-ID`.
- `--site` is validated against `lever.CompaniesBySite`, same allowlist policy as `cmd/workday --tenant`. All curated sites are on the global instance, so there is no `--eu` flag (gated on an EU company entering `companies.yaml`).
- Much simpler than the Workday CLI: Lever's list response already carries full posting content (`descriptionPlain`), so there is no per-result detail fan-out, no errgroup, and no html2text — output renders through a flat `postingJSON` projection shared by text and JSON modes.

Design spec and implementation plan included under `docs/superpowers/`.

## Test plan

- [x] `go test -race -vet=all ./...` — all packages pass (8 CLI tests: 3 rendering-projection unit tests + 5 offline validation error paths)
- [x] Live smoke tests: `companies`, `search --limit 2`, `search --format json`, `get <id>`, 404 on unknown posting id, unknown/missing `--site`, missing subcommand — all behave as spec'd with exit 1 on errors


amikai:
Folding these commits into #80 directly; the CLI ships with the provider in one PR.

PR #107, opened 2026-07-10
test(go): standardize on testify assert/require across the suite

amikai:
## Summary
- Converted the remaining plain-`testing` test files to `stretchr/testify` (`assert`/`require`) for consistency with the rest of the suite: `internal/ats/{filter,registry,greenhouse,ashby,lever,workday}_test.go`, `internal/openingsmcp/company_test.go`, `internal/provider/workday/path_test.go`.
- Fixed genuine `testifylint` findings in already-testify files:
  - `require.ErrorContains` for the first of two chained error assertions on the same `err` in `cmd/{ashby,greenhouse,lever}/main_test.go`, so a mismatched message fails fast instead of a second, less-useful failure.
  - Replaced `require.NoError` with `assert.NoError` (+ early return) inside `httptest` handler goroutines in `internal/provider/{cake,nvidia}/client_test.go`, since `require`'s `t.FailNow()` is unsafe outside the test's own goroutine.
  - `internal/logging/logging_test.go`: preconditions that gate further assertions are now `require` instead of `assert`.
- Along the way, fixed a real bug: `TestAshbySearchIsDeterministic` was silently discarding both `Search` call errors (`_`); it now asserts `NoError` on each.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./...`
- [x] `golangci-lint run --enable-only testifylint ./...` → 0 issues


amikai (commented):
Reviewed the full diff. No actionable findings: the assertion conversions preserve coverage, the require/assert choices are appropriate (including the httptest handler goroutines), and the Ashby test now checks both Search errors. Verified on the PR head: go build ./..., go vet ./..., go test -race ./..., and golangci-lint run --enable-only testifylint ./... (0 issues).

PR #112, opened 2026-07-10
feat(ats): accept careers-page URLs in the unified company tools

amikai:
The unified company tools' `company` parameter now also accepts a careers-page URL on Workday, Greenhouse, Lever, or Ashby, so companies outside the curated roster are queryable with no registration. Stateless: config is re-derived from the URL on every call.

- `workday.CareersSite`: parse `<tenant>.<wd*>.myworkdayjobs.com/<site>`, locale prefixes and deep links stripped. myworkdaysite.com is deliberately not supported — its tenants are all reachable via myworkdayjobs.com (#113)
- Adapters gain `ParseCareersURL`; workday uses the canonical URL as a self-describing slug for non-roster tenants, the slug-family adapters return the board token / org
- `Registry.Resolve` polls adapters for URL-shaped input after the name/slug maps miss; URL misses get their own teaching error
- Detail's company name falls back to the slug for non-roster boards
- MCP schema text and server instructions teach the URL affordance

Spec: docs/superpowers/specs/2026-07-10-careers-url-resolve-design.md

Live-verified against a non-roster Workday tenant (salesforce.wd12) end to end.

## How it fits together

Startup — each provider's curated companies.yaml is projected by `Roster()` and unioned into the registry's two lookup maps; cross-adapter collisions fail startup:

registry build flow

Per call — `Resolve` tries slug, then display name, then (for URL-shaped input) polls each adapter's `ParseCareersURL`; every path converges on `(Adapter, slug)`, and full misses get a teaching error instead of suggestions computed against a URL:

resolve hit flow

After the hit — the slug's meaning splits by adapter family: Workday exchanges it for tenant config (roster map first, self-describing URL slug second) to hit the CXS API server-side, while Greenhouse/Lever/Ashby put it straight into the API path and filter the full board dump locally:

slug consumption


amikai:
**Bottom line:** design is sound and the myworkdayjobs.com / Greenhouse / Lever / Ashby paths are correct and well tested (all tests pass locally), but the myworkdaysite.com support doesn't match how that domain actually works — every real myworkdaysite URL gets rejected. One stale doc comment should also go out with this PR.

### 1. myworkdaysite.com URLs never parse (correctness)

`internal/provider/workday/careers_url.go:30-38` requires exactly four host labels with the tenant first (`<tenant>.<wd*>.myworkdaysite.com/<site>`). Real myworkdaysite.com URLs put the tenant in the **path**, on a three-label host:

```
https://wd<N>.myworkdaysite.com/<locale?>/recruiting/<tenant>/<site>
```

Evidence:

- This repo's own `internal/provider/workday/sp500_tenants.json` records Devon Energy at `wd5.myworkdaysite.com/en-US/recruiting/devonenergy/Careers`, with the actual site being `Careers`.
- Live examples all match that shape: [Devon Energy](https://wd5.myworkdaysite.com/en-US/recruiting/devonenergy/Careers), [Wells Fargo](https://wd1.myworkdaysite.com/recruiting/wf/WellsFargoJobs), [Univ. of Washington](https://wd5.myworkdaysite.com/recruiting/uw/UWHires), [GFL](https://wd3.myworkdaysite.com/recruiting/gflenv/Careers).
- The four-label form the test invents (`acme.wd1.myworkdaysite.com`) doesn't appear to exist — `devonenergy.wd5.myworkdaysite.com` doesn't resolve in DNS.

Consequences: a real myworkdaysite URL fails the `len(labels) != 4` check and gets the "unrecognized careers URL" teaching error (fails closed, at least). Even if a four-label host existed, `segs[0]` would extract `recruiting` as the site — which is exactly what the test case at `careers_url_test.go` ("myworkdaysite host kept verbatim") encodes. The doc comments, spec, and PR description claim myworkdaysite support that can't trigger. (The live verification target, salesforce.wd12, is myworkday**jobs**.com — consistent with only that family working.)

Two coherent fixes:

- **Implement the real shape:** accept three-label `wd*.myworkdaysite.com` hosts, strip the optional locale, then require `recruiting/<tenant>/<site>`; CXS base stays `https://<host>/wday/cxs/<tenant>/<site>`, canonical slug becomes `https://<host>/recruiting/<tenant>/<site>` (round-trip still holds).
- **Or drop myworkdaysite from this PR** — remove it from the domain check, comments, and test, keeping scope myworkdayjobs-only. Smaller, and the Resolve teaching error already only advertises myworkdayjobs.

### 2. Stale invariant comment

`internal/ats/ats.go:27-30` still says "a slug that reaches an adapter is always one it declared" — contradicted five lines later by the new `ParseCareersURL` doc, and the spec explicitly names this comment as the invariant being relaxed. It should now say slugs come from `Roster()` or `ParseCareersURL`.

### Nit

- `internal/ats/workday.go:177`: `strings.ToLower(site.Tenant)` is redundant — `workday.ParseCareersURL` lowercases the whole host before splitting.

### What's good

- Stateless re-derivation keeps the MCP interface unchanged; the canonical-URL-as-slug trick for Workday is well justified and guarded by a round-trip test.
- Error paths are deliberate and tested: URL-shaped misses get a teaching error instead of meaningless Levenshtein suggestions.
- No SSRF concern: the user-controlled host only becomes a request target on the Workday path and is constrained to Workday-owned domains; the other adapters keep fixed hosts.
- URL branch runs only after both map lookups miss; `dump` was already fetch-per-call, so non-roster slugs add no cache-growth vector.
- Coverage is thorough for the shipped behavior: per-adapter host tables, scheme-less input, roster fold-back, Detail company-name fallbacks on all three slug adapters, offline teaching-error tests.

amikai:
Addressed in 3c38fe5:

- **Point 1 (myworkdaysite never parses):** confirmed — the tenant-in-path shape is real (verified Devon Energy / Wells Fargo / GFL live, plus CXS probe; the four-label host is NXDOMAIN). Not fixing in this PR: tracked in #113 with examples and open questions, and the code now carries a KNOWN ISSUE note on ParseCareersURL plus a test case pinning the current rejection.
- **Point 2 (stale invariant comment):** reworded — slugs now come from Roster() or ParseCareersURL.
- **Nit (redundant ToLower):** dropped; the provider parse already lowercases the host.

PR #115, opened 2026-07-10
fix(ats): harden roster and search behavior

amikai:
## Context

The unified ATS package had several correctness gaps found during a package-wide review:

- Workday roster caching stopped deduplicating shared tenants, causing registry construction and server startup to fail for the duplicate `fox` and `dowjones` rows.
- Extreme page values could overflow pagination arithmetic, panicking full-board searches or producing invalid Workday offsets.
- Lever search indexed only `descriptionPlain`, omitting requirements and other content stored in `lists` and `additionalPlain`.
- Full-board ranking collapsed department/team matches and JD-body matches into the same tier.

## Changes

- Keep the cached Workday roster deduplicated, lowercase, and isolated from caller mutation.
- Make full-board pagination overflow-safe and reject unrepresentable Workday offsets with actionable guidance.
- Index the complete Lever job description.
- Rank whole-query matches by title, then organization unit, then combined/JD content.
- Add regression coverage for duplicate tenants, registry construction, cache isolation, extreme pages, complete Lever descriptions, ranking tiers, and deterministic ID tie-breaking.

## Impact

Registry startup succeeds with the curated Workday data, extreme pagination inputs no longer panic or generate invalid offsets, and Lever/full-board searches return more complete and relevant results.

## Validation

- `go test ./... -count=1`
- `go test -race ./internal/ats -count=1`
- `go vet ./internal/ats`
- Independent code review with no remaining actionable findings


amikai:
## Review

**LGTM.** The fixes are correct and each behavior change has a regression test. Verified `go test ./internal/ats` and `go vet` pass at the PR head (494d946) in a clean worktree.

### Correctness

- **`searchDump` pagination** — traced the arithmetic: `pageIndex * PageSize` is only computed when `pageIndex <= total/PageSize`, which bounds the product at `total`, and `end := start + min(PageSize, total-start)` can't exceed `total`. Edge cases (empty set, exact page multiple, `Page: math.MaxInt`, negative page via `clampPage`) all resolve correctly.
- **Workday guard** — `pageIndex > math.MaxInt/PageSize` correctly bounds the offset, and moving the computation ahead of `probeFacets` means a bad page fails before any upstream request (the test asserts zero requests were sent). The asymmetry with `searchDump` — error here vs. a truthful empty page there — is right: the dump engine knows the real total, Workday can't represent the offset at all.
- **Ranking** — tier 1 correctly lowercases `orgUnit` (Lever stores it raw-case), and since `matchQuery`'s blob already includes `orgUnit`, every rank-1 job is a genuine match.
- **Lever descriptions** — reuses the `leverDescription` helper `GetJob` already used, so search and detail now index/display the same text. A conversion error in one posting fails the whole dump, matching `GetJob`'s existing behavior — consistent, and `html2text` errors are effectively unreachable in practice.

### Tests

Good coverage: both overflow guards, org-unit-beats-body ranking, ID tie-break at equal rank+time, `lists`/`additionalPlain` content matching, and roster lowercase/uniqueness/registry-construction/slice-independence. The `bodies` assertion in `TestWorkdaySearchRejectsHugePage` is a nice touch — it pins the fail-fast property, not just the error.

### One note — PR description is stale on the roster

The body's first bullet says "Keep the **cached** Workday roster deduplicated… and isolated from caller mutation," but the net diff contains no roster implementation change: c00a241 added roster caching, then 494d946 removed it, and the resulting `Roster()` is byte-identical to main (which already deduplicates and lowercases). What this PR actually adds for the roster is the regression tests — including `TestWorkdayRosterReturnsIndependentSlice`, which currently passes trivially but guards against reintroducing a shared cached slice. Worth a one-line edit to the PR body so it doesn't describe removed code.

### Nitpick (optional)

- `filter.go` pagination guard could lean on the existing helper — `start := total; if pageIndex < totalPages(total) { start = pageIndex * PageSize }` — which reads as "does this page exist" and is equally overflow-safe. Purely a readability preference; the current form is correct.

PR #124, opened 2026-07-11
fix: mark provider response schemas nullable to stop upstream null decode crashes

### LINKED ISSUE #123: Greenhouse adapter: null Office.location / metadata.value breaks search & detail for most companies
## Summary

Running the unified `search_jobs_by_company` MCP tool against 30 sampled Greenhouse companies, 17/30 failed outright, and 4 more failed on job detail fetch. Root cause is the generated schema being too narrow for fields Greenhouse's real API sends as `null`.

## Evidence

Search failures (identical shape across all 17):

```
greenhouse: list jobs for "discord": decode response: decode application/json: decode JobListResponse:
callback: decode field "jobs": callback: decode JobSummary: callback: decode field "offices":
callback: decode Office: callback: decode field "location": unexpected byte 110 'n' at 7776
```

Affected in the sample: Discord, Safari AI, Hume AI, Datadog, Chime, DoorDash, Asana, Arize AI,
EQT Corporation, Stability AI, Dropbox, Wayve, Contentful, Amplemarket, Gusto, Scandit, Instacart.

Detail-fetch failures (companies whose search passed, but `get_job_detail_by_company` on a real
result failed):

```
greenhouse: fetch job "8001339" for "pinterest": decode response: decode application/json: decode JobDetail:
callback: decode field "metadata": callback: decode JobDetailMetadataItem: callback: decode field "value":
unexpected byte 91 '[' at 591
```

Affected: Pinterest, Hootsuite, Duolingo, Databricks — some with `value` as `null`, some as a JSON array.

## Root cause

`internal/provider/greenhouse/openapi.yaml:104-112` declares `Office.location` as a plain `string`:

```yaml
Office:
  type: object
  properties:
    id:
      type: integer
    name:
      type: string
    location:
      type: string
```

Codegen produces `Location OptString` (`internal/provider/greenhouse/oas_schemas_gen.go:1028`) — absent-or-string
only. Greenhouse's real API returns `"location": null` for some offices, which the generated decoder can't
accept (fails on the literal `null`). Since one bad `Office` fails the whole `jobs` array decode, any company
with at least one such office loses its entire search result — over half the sampled companies.

The `JobDetailMetadataItem.value` field has the same class of bug: it's typed too narrowly for the `null` and
array values Greenhouse's API actually sends for custom metadata fields.

## Fix

Mark `location` nullable (and correct `metadata[].value`'s type) in
`internal/provider/greenhouse/openapi.yaml`, then regenerate the client so codegen produces a proper
nullable/union type instead of a bare `string`.

## Test data used

30 companies sampled from `internal/provider/greenhouse/companies.yaml`, tested live through the real
MCP server (`search_jobs_by_company` / `get_job_detail_by_company`), not mocked.

amikai:
## Summary

Fixes #123. That issue reported Greenhouse's `search_jobs_by_company` failing outright for 17/30 sampled companies because `Office.location` was declared as a plain (non-nullable) string, and upstream sometimes sends `null` there — one bad field failed the entire response decode.

Rather than patch just those two fields, this applies the same fix across every ogen-generated provider: for each schema, non-identifier response fields get `nullable: true` added; identifier fields (job IDs, slugs — anything fed back into a follow-up API call) are deliberately left strict, since a `null` there is a genuine data problem that should keep failing loudly.

- **Greenhouse**: `Office.location` (the original crash), plus the rest of `JobSummary`/`JobDetail`/`Location`/`Department`/`Office`. Also loosened `JobDetail.metadata[].value` to untyped — it's unconsumed and real payloads have shown both `null` and array values there.
- **Workday, Lever, Ashby, Cake, 104, NVIDIA**: same treatment, scoped per provider's actual schema and identifier field(s).
- **Google, LinkedIn, TSMC**: confirmed no `components.schemas` at all (pure HTML-scraping providers, nothing to harden here).
- **Synopsys**: confirmed unregistered — not reachable via any MCP tool, nothing to fix.
- Ashby's `openapi.yaml` was missing from the Makefile's `validate-openapi` list; added it.

No adapter/tool Go logic changes were needed for most of this — ogen's `OptString`→`OptNilString` (and `string`→`NilString` for required-but-nullable fields) preserves the same `.Value` accessor, so most existing call sites kept compiling as-is. Where a field was *required* (bare, not `Opt`-wrapped) and became nullable, a handful of call sites needed `.Value` added — those are the only non-schema code changes in this branch.

## Test plan

- [x] `go build ./...` and `go test ./...` pass for all seven touched providers
- [x] Real-upstream smoke checks (search + detail) for Cake, 104, and NVIDIA — not part of the full-roster sweep below, verified separately against live data
- [x] Full-roster live verification: ran `search_jobs_by_company` (+ `get_job_detail_by_company` spot-check) through the real MCP server against all 360 curated companies across Workday/Lever/Ashby/Greenhouse — **351/360 passed, zero decode/nullable-related failures**. The 9 remaining failures are pre-existing and unrelated (8 are a roster-file sync gap between `sp500_workday_confirmed.json` and `companies.yaml`; 1 is RunPod's Greenhouse board 404ing upstream) — details in the [issue #123 comment](https://github.com/amikai/openings-mcp/issues/123#issuecomment-4947534751), left open to track those separately.

Design and plan docs for this work are included under `docs/superpowers/`.


internal/provider/greenhouse/openapi.yaml
```diff
@@ -129,14 +133,18 @@ components:
           description: Null for prospect posts.
         title:
           type: string
+          nullable: true
```
amikai, line 136: **[P1] Please update the Greenhouse CLI tests for the regenerated nullable types.** `go test ./...` currently fails to compile `cmd/greenhouse/main_test.go`: lines 24–28 and 41 still use `NewOptString`, `NewOptDateTime`, and `NewOptURI`, while these fields now require the `OptNil` constructors. This is also why the PR's CI run is failing.

internal/ats/greenhouse.go
```diff
@@ -190,7 +190,7 @@ func greenhouseDescription(content string) string {
 	return text
 }
 
-func greenhousePostedAt(t greenhouse.OptDateTime) string {
+func greenhousePostedAt(t greenhouse.OptNilDateTime) string {
```
amikai, line 193: **[P2] Handle present-but-null timestamps here.** `OptNilDateTime.Set` is true when the JSON field is explicitly `null`, so the existing `!t.Set` guard falls through and formats the zero time as `0001-01-01`. Please use `Get()` or check `t.Null` as well. The `cmd/greenhouse` `FirstPublished`/`UpdatedAt` paths have the same Set-only assumption.

internal/ats/ashby.go
```diff
@@ -76,11 +76,11 @@ func (a *AshbyAdapter) Detail(ctx context.Context, slug, jobID string) (*JobDeta
 		}
 		return &JobDetail{
 			JobID:       j.ID.Value,
-			Title:       j.Title,
+			Title:       j.Title.Value,
 			Company:     cmp.Or(ashby.CompaniesByBoard[slug].Name, slug),
 			Location:    ashbyLocations(j),
-			PostedAt:    isoDate(j.PublishedAt),
-			URL:         j.JobUrl,
+			PostedAt:    isoDate(j.PublishedAt.Value),
```
amikai: **[P2] Preserve a null `publishedAt` instead of emitting a fake date.** Since `PublishedAt` is now nullable, reading `.Value` directly turns null into the zero time and `isoDate` returns `0001-01-01`. The dump path and `cmd/ashby.summarize` repeat this pattern; guard `Null`/use `Get()` and leave the date empty when unknown.

internal/provider/lever/openapi.yaml
```diff
@@ -202,11 +203,13 @@ components:
         createdAt:
           type: integer
           format: int64
+          nullable: true
```
amikai, line 206: **[P2] Update consumers to distinguish null from a real epoch value.** `internal/ats/lever.go:171` and `cmd/lever/main.go:281` only check `.Set`; for explicit null that is true and `.Value == 0`, so users receive `1970-01-01`. Those paths should use `Get()` or check `Null` before formatting.

internal/openingsmcp/job104.go
```diff
@@ -290,28 +290,28 @@ func job104HTTPToMCPResponse(resp *job104.JobsResponse) *job104SearchOutput {
...
+			CompanyURL:    j.Link.Cust.Value,
+			SalaryHigh:    j.SalaryHigh.Value,
+			SalaryLow:     j.SalaryLow.Value,
+			SalaryType:    job104SalaryTypeLabels[j.S10.Value],
+			JobAddrNoDesc: j.JobAddrNoDesc.Value,
+			AppearDate:    j.AppearDate.Value,
+			ApplyCnt:      j.ApplyCnt.Value,
+			Remote:        job104RemoteWorkLabels[job104.SearchJobsRemoteWork(j.RemoteWorkType.Value)],
+			JobType:       job104RoLabels[job104.SearchJobsRo(j.JobRo.Value)],
+			Experience:    job104ExperienceLabel(j.Period.Value),
```
amikai: **[P2] Do not classify a null period as `Under1Year`.** A null `NilInt` has `.Value == 0`, and `job104ExperienceLabel(0)` returns `Under1Year`, turning unknown upstream data into a concrete claim. Check `Get()`/`Null` and omit or leave the experience value empty when the period is null.

internal/provider/workday/openapi.yaml
```diff
@@ -262,10 +263,12 @@ components:
       properties:
         facetParameter:
           type: string
+          nullable: true
```
amikai, line 266: **[P2] Audit the Set-only facet consumers after making this nullable.** In `internal/ats/workday.go:265`, a present-but-null `facetParameter` has `Set == true`, so its empty `.Value` overwrites the parameter inherited from the parent and valid leaf facets are dropped. `cmd/workday/main.go` has the same Set-only interpretation. Use `Get()` so only a non-null value changes the facet parameter.

amikai (commented):
Inline review findings from local build and nullable-semantics audit.

internal/provider/greenhouse/openapi.yaml
amikai: Fixed in cmd/greenhouse/main_test.go — updated to the OptNil/NilDateTime/NilURI constructors (e506d5f).

internal/ats/greenhouse.go
amikai: Fixed: greenhousePostedAt now uses Get() instead of !Set, so a present-but-null FirstPublished/UpdatedAt no longer formats the zero time as 0001-01-01. Same fix applied to the cmd/greenhouse call sites. Added TestSummarizeNullDates (e506d5f).

internal/ats/ashby.go
amikai: Fixed: added an ashbyPostedAt helper using Get() instead of reading .Value directly, used in both Detail and dump. Fixed the matching cmd/ashby/main.go summarize() call site too. Added TestAshbyPostedAtNull (e506d5f).

internal/provider/lever/openapi.yaml
amikai: Fixed: leverPostedAt now uses Get() instead of .Set, so a present-but-null createdAt no longer formats epoch 0 as 1970-01-01. Same fix applied to cmd/lever/main.go's toPostingJSON. Added TestLeverPostedAtNull (e506d5f).

internal/openingsmcp/job104.go
amikai: Fixed: job104ExperienceLabel now takes the NilInt directly and returns "" on null instead of conflating null with a real period=0 ("no requirement"). Added a null case to TestJob104ExperienceLabel (e506d5f).

internal/provider/workday/openapi.yaml
amikai: Fixed: flattenFacets now uses Get() for both facetParameter and descriptor, so a null facetParameter no longer overwrites the parent's inherited param, and a null descriptor is dropped instead of emitted as an empty-label entry. Same fix applied to cmd/workday/main.go's printFacetNode. Added TestFlattenFacetsNullFacetParameter (e506d5f).

PR #131, opened 2026-07-11
Add shared 1GiB HTTP response cache across all providers

amikai:
Every tool call re-fetched upstream in full. The ATS boards are the worst case: no server-side search, so each search/filters call re-downloads the whole board, and Ashby's detail does too (no per-job endpoint) — one real conversation pulled OpenAI's 12MB board ten times at ~20s each. The dumps are parameterless snapshots that change on an hours scale, so caching at the HTTP layer is safe and covers every provider at once.

- `internal/httpcache`: otter v2 RoundTripper cache — 1GiB shared byte budget (weigher = key+body+headers), 60min write-expiry, key = method+URL+sha256(body), Cookie/Authorization excluded, GET/POST 2xx only, hop-by-hop headers stripped
- `cmd/openings-mcp`: one cache instance wraps all three http.Client transports

Measured end-to-end: repeated identical `search_jobs_by_company anthropic` goes 1.45s → 190ms (remaining cost is JSON decode + html2text, a known trade-off; see `docs/superpowers/specs/2026-07-12-httpcache-design.md`). Deliberately out of scope: ETag revalidation, singleflight — concurrent identical requests each fetch once, harmless for sequential LLM usage.


internal/httpcache/httpcache.go
```diff
@@ -0,0 +1,158 @@
...
+// New returns a Cache holding at most maxBytes of responses, each entry
+// expiring ttl after it was stored. A nil logger disables hit/miss logging.
+func New(maxBytes uint64, ttl time.Duration, logger *slog.Logger) *Cache {
+	if logger == nil {
+		logger = slog.New(slog.DiscardHandler)
+	}
+	return &Cache{
+		store: otter.Must(&otter.Options[string, *entry]{
+			MaximumWeight:    maxBytes,
+			Weigher:          weigh,
```
amikai, line 47: Otter starts a periodic cleanup goroutine whenever `ExpiryCalculator` is configured. `Cache` exposes no shutdown method, so each `New` leaves that goroutine running after the server or a test ends. Please add `Close` (calling `store.StopAllGoroutines()`), defer it in `runWithTransport`, and register it with `t.Cleanup` in the package tests.

amikai (commented):
One blocking cache-key correctness finding.

internal/httpcache/httpcache.go
```diff
@@ -0,0 +1,158 @@
...
+	if req.Body != nil {
+		b, err := io.ReadAll(req.Body)
+		req.Body.Close()
+		if err != nil {
+			return "", fmt.Errorf("httpcache: read request body: %w", err)
+		}
+		sum.Write(b)
+		req.Body = io.NopCloser(bytes.NewReader(b))
+	}
+	return req.Method + " " + req.URL.String() + " " + hex.EncodeToString(sum.Sum(nil)), nil
```
amikai, line 139: **[P1] Include response-varying headers in cache keys**

The key omits every request header, but Indeed carries the selected country only in `indeed-co` (`internal/provider/indeed/client.go:62-64`); equivalent searches against different countries therefore have the same URL and body. After a Taiwan search, an equivalent US search can reuse Taiwan’s response for 60 minutes, after which `Jobs` relabels and relinks that payload as US data. Please include `indeed-co`—ideally a canonical set of representation-affecting headers—in the key and add a regression test with consecutive searches for two countries.

internal/httpcache/httpcache.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if e, ok := t.cache.store.GetIfPresent(key); ok {
+		t.cache.logger.Debug("httpcache hit", "key", key, "size", len(e.body))
+		return e.response(req), nil
+	}
+	resp, err := t.next.RoundTrip(req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
```
amikai, line 106: **[P1] Do not cache session-bootstrap responses**

This stores every 2xx response despite its cache directives. Apple calls `/api/v1/CSRFToken` before each search; that endpoint returns a unique CSRF token and `jssid`, plus `Cache-Control: no-store, no-cache`. Replaying the cached response suppresses session renewal, and `jssid` expires after 30 minutes while this cache retains the response for 60, so later uncached search bodies can be sent with an expired session. Please bypass caching for `no-store`/`no-cache` responses (and preferably responses containing `Set-Cookie` while `Cookie` remains excluded from the key), with a rotating-session regression test.

amikai (commented):
One remaining cacheability blocker involving session bootstrap responses.

PR #150, opened 2026-07-12
feat: add SmartRecruiters ATS adapter and wire it into the MCP server

amikai:
Completes the SmartRecruiters integration (stage 6 of the provider pipeline): the 52 rostered companies — and any jobs.smartrecruiters.com careers URL — now join the unified company tools. Location folds into the API's \`q\` param since there is no cheap way to enumerate a tenant's locations; \`department\` labels resolve to ids via one departments call (comma-joined ids OR, verified live). Spec and plan under docs/superpowers/.

Verified live: \`verify-companies --provider smartrecruiters\` passes 52/52, plus a Filters/Search/Detail smoke against Equinox.


internal/ats/smartrecruiters.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	params := smartrecruiters.ListPostingsParams{
+		CompanyIdentifier: slug,
+		Limit:             smartrecruiters.NewOptInt(PageSize),
+		Offset:            smartrecruiters.NewOptInt(pageIndex * PageSize),
+	}
+	// q full-text matches titles and location text upstream, so the
+	// unified Location folds into it rather than guessing among the
+	// exact-match country/region/city params.
+	if q := strings.TrimSpace(strings.TrimSpace(p.Query) + " " + strings.TrimSpace(p.Location)); q != "" {
```
amikai: SmartRecruiters `q` searches both titles and locations and combines terms with OR semantics. A live Equinox check returned 138 results for `trainer`, 9 for `Houston`, but 146 for `trainer Houston`; therefore `SearchParams{Query: "trainer", Location: "Houston"}` broadens the search instead of requiring both constraints. It also allows `Query` alone to match locations, contrary to the unified contract. Keep these dimensions separate or use a pagination-safe residual filter, and add a behavioral test—the current test only verifies the outgoing query string.

internal/ats/smartrecruiters.go
```diff
@@ -0,0 +1,502 @@
...
+}
+
+func (a *SmartRecruitersAdapter) chooseSmartRecruitersCandidates(
+	ctx context.Context,
+	slug, query, location string,
+	base smartrecruiters.ListPostingsParams,
+) (string, *smartrecruiters.PostingList, error) {
+	seed := query
+	if seed == "" {
+		seed = location
```
amikai, line 169: `Location: "remote"` silently under-returns when it becomes the candidate seed (Query empty, or the location set is smaller and wins below). Upstream `q` matches title/location *text* only, not the `remote` flag, so `matchLocation`'s `isRemote` branch only ever sees postings that already contain the word "remote".

Confirmed live against Delivery Hero (in the roster): `locationType=REMOTE` → **14** postings, all with city-only location text (`"Angeles, Central Luzon, Philippines"`, `remote: true`, …), while `q=remote` → **6**. So `search_jobs_by_company(company: "Delivery Hero", location: "remote")` misses most of the actual remote jobs — and per the `SearchParams` doc, `location: "remote"` is exactly what full-dump adapters honor via their remote fields.

Suggestion: special-case `location == "remote"` to filter server-side with `locationType=REMOTE` (already wired up for the `location_type` filter) instead of using it as a `q` seed. The asymmetric case is fine as-is: when Query is non-empty and wins as seed, the local `isRemote` check works on the full query candidate set.

internal/ats/smartrecruiters.go
```diff
@@ -0,0 +1,502 @@
...
+		return nil, fmt.Errorf("smartrecruiters: job %q not found for company %q; pass a job_id exactly as returned by the job search", jobID, slug)
+	}
+	_, name := resolveSmartRecruitersCompany(slug)
+	return &JobDetail{
+		JobID:       cmp.Or(d.ID.Value, jobID),
+		Title:       d.Name.Value,
+		Company:     cmp.Or(d.Company.Value.Name.Value, name),
+		Location:    d.Location.Value.FullLocation.Value,
+		PostedAt:    smartRecruitersPostedAt(d.ReleasedDate),
+		URL:         d.PostingUrl.Value,
```
amikai: Minor: the schema doesn't mark `postingUrl` required, so this can come back empty while search summaries always derive a URL. `cmp.Or(d.PostingUrl.Value, smartRecruitersPostingURL(identifier, jobID))` would match the defensive `cmp.Or` treatment the surrounding fields already get.

amikai (commented):
Reviewed the adapter; verified the design's key assumptions live against the public API. One correctness gap (inline), one minor hardening nit (inline). Everything else looks merge-ready — the fix is small since `LocationType` plumbing already exists.

PR #151, opened 2026-07-13
Add Teamtailor provider and CLI support for the new ATS adapter

amikai:
## Summary
- Adds a new Teamtailor provider and ATS adapter so Openings MCP can surface jobs from Teamtailor-backed companies.
- Introduces the Teamtailor CLI and verification entrypoints, plus wiring in the main binary and registry.
- Updates generated OpenAPI server/client code, test fixtures, docs, and roster data to support the new provider.

## Testing
- Added and updated unit tests for the Teamtailor provider, adapter, and CLI entrypoints.
- Validated the new provider behavior with request/response fixtures covering normal, null, and unknown-company cases.
- Not run (not requested)


internal/provider/teamtailor/mocksrv.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/http/httptest"
+)
+
+// MockHost is the live career-site host whose response is captured in
+// testdata/jobs_rsp.json.
+const MockHost = "knaufsemea.teamtailor.com"
+
+// MockNonRosterHost is deliberately absent from companies.yaml so ATS tests
+// can exercise a URL-resolved Teamtailor tenant outside the curated roster.
+const MockNonRosterHost = "somestartup.teamtailor.com"
```
amikai, line 15: `MockNonRosterHost` is unused — the comment says ATS tests use it, but nothing references it. Remove it, or add the non-roster-host test the comment promises.

internal/ats/teamtailor.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/jaytaylor/html2text"
+
+	"github.com/amikai/openings-mcp/internal/provider/teamtailor"
+)
+
+var _ Adapter = (*TeamtailorAdapter)(nil)
+
+// TeamtailorAdapter serves Teamtailor career sites. The public /jobs.json
+// endpoint returns the complete board with full descriptions, so all search,
+// filter, and detail behavior is implemented over that dump.
```
amikai, line 20: The feed carries no remote-work signal, so `dumpJob.isRemote` is never set and `location: "remote"` finds nothing even on remote-first boards — confirmed live on curated bunny.net, whose fully-remote postings list only office addresses. Nothing to fix in code (the data isn't in the feed), but worth a sentence here so the limitation is recorded rather than rediscovered.

cmd/teamtailor/main.go
```diff
@@ -0,0 +1,290 @@
...
+	feed, err := fetchFeed(ctx, host, timeout)
+	if err != nil {
+		return err
+	}
+
+	for _, j := range feed.Items {
+		if j.ID.String() != jobID {
+			continue
+		}
+		description, err := html2text.FromString(j.ContentHTML, html2text.Options{})
```
amikai, line 262: On `html2text` failure the CLI errors out, while the adapter (`teamtailorDescription`) falls back to raw HTML. Same input, two behaviors — the fallback seems right for both.

amikai (commented):
Reviewed the adapter, provider, and CLI. No blocking issues — three minor inline notes. Schema strictness was spot-checked against live feeds (4 tenants incl. bunny.net and the NA region host): every item carries `_jobposting.jobLocation` with non-null locality/country, so the required-field spec holds.

PR #163, opened 2026-07-13
feat(provider): integrate Recruitee ATS provider

amikai:
Adds Recruitee as a new ATS provider, following the existing full-dump adapter pattern (Teamtailor, Lever): fetch the whole board once, then search/filter/detail client-side.

- Hand-written OpenAPI 3.1.0 spec for Recruitee's public `/api/offers` feed, with ogen-generated client
- `RecruiteeAdapter` implementing the unified ATS interface (search, filters, detail, careers-URL parsing) via `searchDump`
- Embedded roster of confirmed companies (Hygraph, Surfly, bunq), each verified live against `/api/offers`
- Mock server + fixtures (including a null-fields fixture) backing client and adapter tests
- `cmd/recruitee` debug CLI for inspecting a board outside the MCP server
- Wired into MCP server main, ATS registry, `verify-companies`, README, and Makefile


cmd/recruitee/main.go
```diff
@@ -0,0 +1,278 @@
...
+	dept := o.Department.Or("")
+	loc := o.Location.Or("")
+
+	jobURL := o.CareersURL.Or("")
+	if jobURL == "" && !o.Slug.Null && o.Slug.Value != "" {
+		jobURL = fmt.Sprintf("https://%s.recruitee.com/o/%s", slug, o.Slug.Value)
+	}
+
+	posted := ""
+	if o.PublishedAt.Set {
```
amikai: `o.PublishedAt.Set` is true even when the API returns an explicit `null` (`Set=true, Null=true`), so `posted` becomes `""` and the `created_at` fallback on the next branch is never consulted. The adapter handles this correctly with `o.PublishedAt.Or(o.CreatedAt.Or(""))` — the same `.Or()` chain works here. Debug-CLI only, so low stakes.

internal/ats/registry.go
```diff
@@ -104,6 +104,7 @@ var careersHostPatternsByAdapter = map[string]string{
 	"lever":      "jobs.lever.co/<org>",
 	"ashby":      "jobs.ashbyhq.com/<org>",
 	"teamtailor": "<company>[.na|.au].teamtailor.com/jobs",
+	"recruitee":  "<company>.recruitee.com/o/<job>",
```
amikai: This pattern advertises a job-posting URL, while every other entry advertises the careers-page root that `ParseCareersURL` accepts (e.g. `jobs.lever.co/<org>`). The Recruitee adapter accepts any path on the subdomain, so `<company>.recruitee.com` is both more accurate and more consistent with its neighbors.

PR #169, opened 2026-07-14
feat: integrate SAP SuccessFactors as a new ATS provider

amikai:
## Summary
- Adds SAP SuccessFactors Career Site Builder as a new multi-tenant ATS provider (`internal/provider/successfactors`, `internal/ats/successfactors.go`, `cmd/successfactors`).
- SuccessFactors has no public developer API for career-site search (the documented OData v2 API requires per-tenant OAuth), so this hand-derives the classic `/search/`, `/job/{slug}/{id}/`, and `/services/jobs/options/facetValues/` endpoints from live traffic — cross-checked against four independent tenants (SAP, adidas, Kellanova, Eastman Chemical) to confirm they're a shared platform feature, not a per-tenant customization.
- Seed roster of 4 companies, each verified live via `cmd/verify-companies`.
- Wired into `newATSRegistry` (`cmd/openings-mcp/main.go`), `careersHostPatternsByAdapter` (`internal/ats/registry.go`), and `providerOrder` (`cmd/verify-companies/main.go`).
- README updated with the new provider.

## Test plan
- [x] `go build ./...`
- [x] `go test -vet=all ./...`
- [x] `make validate-openapi`
- [x] `make hurl-lint`
- [x] `make hurl-test` (live replay against jobs.sap.com)
- [x] `go run ./cmd/verify-companies --provider successfactors` — all 4 roster entries OK
- [x] Manual smoke test of `cmd/successfactors` (search/detail/facets/companies) against the live SAP career site


internal/ats/successfactors.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	page := clampPage(p.Page)
+	startRow := (page - 1) * pageSize
+
+	client := successfactors.NewClient(a.baseURL(c.Host), a.hc)
+	res, err := client.Search(ctx, &successfactors.SearchRequest{
+		Query:          p.Query,
+		LocationSearch: strings.TrimSpace(p.Location),
+		Department:     filterValues["department"],
```
amikai: **[P1] Tenant-defined filters are silently ignored.** `Filters()` advertises every facet returned by the tenant, and `resolveFilters` accepts them, but `Search` forwards only `department`, `customfield3`, and `country`. Live checks found adidas facets such as `city`, `dept`, `shifttype`, and `customfield2`, plus several Swiss Re custom fields. Searches using those advertised filters therefore return unfiltered results without an error. Please build the `optionsFacetsDD_<dimension>` parameters dynamically from the resolved map.

internal/ats/successfactors.go
```diff
@@ -0,0 +1,266 @@
...
+		if !ok || len(options) == 0 {
+			return nil, errUnknownFilterKey(key, valid)
+		}
+		if len(values) == 0 {
+			continue
+		}
+		// Each optionsFacetsDD_<dimension> query param is a single-select
+		// dropdown upstream (see openapi.yaml), so FilterSet's usual OR
+		// semantics across values within one key can't be expressed here.
+		if len(values) > 1 {
```
amikai: **[P2] Preserve the unified OR-filter contract.** `SearchParams` documents OR semantics for multiple values within one filter key, but this adapter rejects that valid input. Since upstream is single-select, the adapter needs to fan out requests and merge/deduplicate results—or otherwise implement the unified contract—instead of exposing provider-specific behavior.

internal/ats/successfactors.go
```diff
@@ -0,0 +1,266 @@
...
+	return toFilterSet(seen), nil
+}
+
+func (a *SuccessFactorsAdapter) Detail(ctx context.Context, slug, jobID string) (*JobDetail, error) {
+	c, err := a.resolveSlug(slug)
+	if err != nil {
+		return nil, err
+	}
+	client := successfactors.NewClient(a.baseURL(c.Host), a.hc)
+	d, err := client.JobDetail(ctx, jobID)
```
amikai, line 288: **[P2] Do not turn every detail failure into not-found.** This path discards the original error and reports timeouts, HTTP 5xx responses, malformed HTML, and cancellation as an expired job. Please distinguish the parser's actual not-found condition with a sentinel or typed error, and wrap operational failures with `%w`.

cmd/successfactors/main.go
```diff
@@ -0,0 +1,336 @@
...
+func main() {
+	rootFlags := ff.NewFlagSet("successfactors")
+	var (
+		company = rootFlags.StringLong("company", "", `curated company name or career-site host, e.g. "SAP" or "jobs.sap.com"`)
+		timeout = rootFlags.DurationLong("timeout", 60*time.Second, "request timeout")
+		format  = rootFlags.StringEnumLong("format", "output format", "text", "json")
+	)
+	rootCmd := &ff.Command{
+		Name:  "successfactors",
+		Usage: "successfactors --company COMPANY [FLAGS] <companies|search|detail> [FLAGS]",
```
amikai: **[P3] Include `facets` in the CLI usage.** The command registers a `facets` subcommand, but the root usage string lists only `companies`, `search`, and `detail`, so the new command is absent from the primary usage synopsis.

amikai (commented):
The implementation and tests are generally solid, but these correctness issues should be addressed before merging. The commit history also needs cleanup: `e7f89e0` and `f054fec` are separate `roster: enrich SuccessFactors companies` commits, contrary to the repository rule requiring a session's roster additions to be batched and squashed.

amikai:
Addressed all review feedback in 7096934:

- **[P1]** `SearchRequest` now carries a generic `Filters map[string]string`; `Search` builds `optionsFacetsDD_<dimension>` for whatever the resolved filter map contains, instead of three hardcoded fields. Verified live against adidas's `city` dimension (1139 → 16 jobs for `city=Auckland`, matching the facet count exactly).
- **[P2]** Multi-value filters now fan out: `filterCombinations` expands OR'd values into the AND'd single-value requests the upstream single-select dropdowns can each express, `searchWithFanout` issues one fully-paginated request per combination and unions+dedupes by job ID, bounded by `maxFilterCombinations`/`maxPagesPerCombination` so a broad selection errors instead of silently returning a partial merge. Covered by a new test using a synthetic two-value mock.
- **[P3]** `JobDetail` now returns `ErrJobNotFound` only for the genuine not-found case; the adapter checks `errors.Is` and wraps everything else (5xx, timeouts, ...) with `%w` instead of relabeling it as an expired job ID. Covered by new tests at both the client and adapter layers.
- **[P4]** Root usage string now lists `facets`.

Also squashed the two `roster:` commits into one per the commit convention, and force-pushed.

internal/ats/successfactors.go
```diff
@@ -0,0 +1,401 @@
...
+// nil-map) combination, so callers don't special-case the no-filter case.
+func filterCombinations(filterValues map[string][]string) []map[string]string {
+	if len(filterValues) == 0 {
+		return []map[string]string{nil}
+	}
+	dims := slices.Sorted(maps.Keys(filterValues))
+
+	combos := []map[string]string{{}}
+	for _, dim := range dims {
+		next := make([]map[string]string, 0, len(combos)*len(filterValues[dim]))
```
amikai: **[P1] Enforce the combination limit before allocation.** `filterCombinations` still materializes the complete Cartesian product before `searchWithFanout` checks `maxFilterCombinations`. Large multi-facet input can allocate hundreds of thousands or millions of maps—and potentially exhaust memory—before returning the intended max-12 error. Compute the product with overflow-safe checks before allocation, and deduplicate repeated values.

internal/ats/successfactors.go
```diff
@@ -0,0 +1,401 @@
...
+			if err != nil {
+				return nil, err
+			}
+			for _, j := range res.Jobs {
+				seen[j.ID] = j
+			}
+			if len(res.Jobs) == 0 || (page+1)*successFactorsUpstreamPageSize >= res.TotalCount {
+				break
+			}
+			if page == maxPagesPerCombination-1 {
```
amikai: **[P2] Do not require downloading the full result set to serve one page.** Any valid filter combination with more than 250 matches still errors even when the caller requested only page 1. Common values such as SAP's `Professional` career status exceed this limit. Fan-out should page lazily or otherwise serve the requested unified page without first fetching every matching job.

amikai (commented):
Rebase is clean and the PR is mergeable again. The checks pass, but these four findings remain. Commit history also needs one cleanup: `923788d` changes `internal/provider/successfactors/companies.yaml` separately from `be6d7e9 roster: enrich SuccessFactors companies`; per the repository convention, those roster changes should be squashed into the roster commit.

internal/provider/successfactors/client.go
```diff
@@ -0,0 +1,227 @@
...
+		return nil, fmt.Errorf("parse base url %q: %w", c.baseURL, err)
+	}
+	u = u.JoinPath("job", id, id)
+	u.Path += "/"
+
+	doc, err := c.getHTML(ctx, u.String())
+	if err != nil {
+		return nil, fmt.Errorf("job detail %q: %w", id, err)
+	}
+	detail, ok := parseJobDetailHTML(doc, id)
```
amikai, line 153: **[P2] Distinguish the actual error page from other HTTP 200 HTML.** Every parseable 200 document without `itemprop="title"` is still classified as `ErrJobNotFound`, including maintenance pages, bot challenges, login pages, and template regressions. Detect the final `/errorpage/` redirect or an explicit error-page marker; return an unexpected-upstream-page error for other HTML.

cmd/successfactors/main.go
```diff
@@ -0,0 +1,345 @@
...
+		return err
+	}
+	if f.startRow < 0 {
+		return fmt.Errorf("--start-row must be >= 0, got %d", f.startRow)
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, f.timeout)
+	defer cancel()
+
+	filters := make(map[string]string, 3)
```
amikai: **[P3] Let the CLI use tenant-defined facets.** The `facets` command can report arbitrary keys such as adidas `city`, `dept`, `shifttype`, and `customfield2`, but CLI search still forwards only three SAP-specific keys. Add a repeatable generic `key=value` filter flag so every facet printed by the CLI can actually be searched.

PR #168, opened 2026-07-13
feat: integrate Eightfold as a new ATS provider

amikai:
## Summary

- Adds full Eightfold ATS support: no public developer docs exist for the career-site search API, so `/api/pcsx/search` and `/api/pcsx/position_details` were reverse-engineered from live browser traffic on morganstanley.eightfold.ai and captured as hurl fixtures (replay-tested live).
- Full pipeline: `openapi.yaml` + generated client, provider package, debug CLI (`cmd/eightfold`), and `internal/ats.EightfoldAdapter` wired into `search_jobs_by_company` / `verify-companies`.
- Two upstream quirks shaped the adapter: the search page size is fixed at 10 (no `limit`/`size` override works), so `Search` fetches two upstream pages per unified 20-job page; and facet filter names (`businessarea`, `employmenttype`, ...) are tenant-specific and sent as `filter_<name>=<value>`, which OpenAPI can't express as a typed parameter — handled by hand in `search.go`/`eightfold.go` instead of the generated client.

## Roster conflicts

Seeded with 5 verified companies: **Eaton, Infineon, Lam Research, Qualcomm, Vodafone**.

Three companies confirmed live on Eightfold were **left out** because their slug or normalized display name already collides with an existing Workday roster entry (`ats.NewRegistry` fails startup on any collision, by design):

| Company | Eightfold tenant | Conflicts with (Workday) |
|---|---|---|
| Micron Technology | `micron` | same slug (`micron`) and same company name |
| Morgan Stanley | `morganstanley` | different slug (`ms`), but company name normalizes to the same key |
| PayPal | `paypal` | same slug (`paypal`) and same company name |

Resolving which provider should own these three is a separate decision outside this PR's scope.

## Test plan

- [x] `go build ./...`
- [x] `go test -race -vet=all ./...`
- [x] `make validate-openapi`
- [x] `make hurl-test` (all 5 fixtures replay live successfully)
- [x] `go run ./cmd/verify-companies --provider eightfold` — all 5 roster companies OK with real job counts
- [x] Manual smoke test of `cmd/eightfold`'s `companies`/`search`/`filters`/`detail` subcommands against the live API

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #171, opened 2026-07-14
fix: merge Eightfold allFilters facets, qualify README roster claim

amikai:
## Summary
- `Filters`/`resolveFilters` only inspected `filterDef.smartFilters`, but Eaton, Infineon, and Qualcomm return empty `smartFilters` with facets populated only in `filterDef.allFilters` instead — 3 of 5 roster tenants exposed no filters. Modeled `allFilters` in the OpenAPI spec (regenerated client) and merged both facet lists.
- README's "pass its careers-page URL on any of those platforms" line included Eightfold, but `ParseCareersURL` intentionally rejects unrostered tenants since the required `domain` value can't be derived from a URL. Qualified the sentence to carve out Eightfold.

Addresses two open review comments on #168:
- https://github.com/amikai/openings-mcp/pull/168#discussion_r3575127676
- https://github.com/amikai/openings-mcp/pull/168#discussion_r3575127678

## Test plan
- [x] `go build ./...`
- [x] `go test ./...`
- [x] Added `TestEightfoldFilters` assertion + `TestEightfoldSearchWithAllFiltersOnlyFacetResolvesLabelToValue` covering an `allFilters`-only facet (`country`) in the mock fixtures


amikai:
Follow-up: tested the previous commit live against eaton/infineon/qualcomm.eightfold.ai (not just the mock fixtures) and found Eaton's and Qualcomm's `latlong_non_remote` facet sends an integer nearby-postings count as `label` instead of a string, which failed decoding the *entire* search response over one unusable facet. Fixed by decoding `allFilters` option labels permissively (`AllFilterOption` schema) and dropping non-string-label options in `normalizeAllFilter`.

Live-verified after the fix:
- Eaton now returns 7 filter dimensions (was a hard decode error)
- `filter_work_type=onsite` drops Eaton's total from 2250 → 1389, matching the review comment's reported numbers (2,253 → 1,394)
- Infineon and Qualcomm both decode and expose filters correctly

internal/ats/eightfold.go
```diff
@@ -163,21 +164,66 @@ func (a *EightfoldAdapter) Filters(ctx context.Context, slug string) (FilterSet,
...
 	}
 	return toFilterSet(seen), nil
 }
 
+// eightfoldFacets concatenates smartFilters and allFilters into one facet
+// list. allFilters is a second, usually-disjoint facet list some tenants
+// (observed: Eaton, Infineon, Qualcomm) populate instead of smartFilters —
+// normalized to the SmartFilter shape so Filters and resolveFilters treat
+// both as one combined set.
+func eightfoldFacets(fd eightfold.FilterDef) []eightfold.SmartFilter {
```
amikai: **[P2] Keep debug CLI filter discovery in sync**

This fixes the unified adapter, but `cmd/eightfold.runFilters` still iterates only `res.Data.FilterDef.SmartFilters` at `cmd/eightfold/main.go:341`. On this PR's head, `go run ./cmd/eightfold --company eaton --format json filters` still prints `null`, so the original `allFilters`-only tenants cannot discover valid `--filter` names and values through the debug CLI. Please centralize or export this normalization for the CLI to reuse (or update the CLI separately) and add an `allFilters`-only CLI regression test.

amikai (commented):
One remaining regression in the Eightfold debug CLI.

amikai:
Fixed: moved the smartFilters/allFilters merge+normalize logic into an exported \`eightfold.MergedFacets\` (internal/provider/eightfold/filters.go), and updated \`cmd/eightfold\`'s \`filters\` subcommand (\`printFilters\`, split out of \`runFilters\` for testability) to call it instead of reading \`SmartFilters\` directly.

Verified live: \`go run ./cmd/eightfold --company eaton --format json filters\` now prints Eaton's 7 facets (work_type, career_level, hiring_program, departments, relocation_assistance_available, organization, sector) instead of \`null\`.

Also added a CLI regression test (\`cmd/eightfold/main_test.go\`) covering an allFilters-only tenant against a mock server, plus moved the normalization unit tests to \`internal/provider/eightfold/filters_test.go\` alongside the new exported function.

PR #174, opened 2026-07-14
feat: add Indeed as a job-board provider

amikai:
## Summary
- Ports jobspy's Python Indeed scraper (`case-study/JobSpy/jobspy/indeed`) to Go: `internal/provider/indeed` (client + GraphQL query builders + country/domain table), `cmd/indeed` (debug CLI), and `internal/openingsmcp/indeed.go` (`indeed_search_jobs` / `indeed_get_job_detail` MCP tools wired into `cmd/openings-mcp`).
- Indeed's search (`jobSearch`) and detail (`jobData(jobKeys: [...])`) are two GraphQL operations on one undocumented endpoint (`apis.indeed.com/graphql`), so the client is hand-written against a documentation-only `openapi.yaml` instead of ogen-generated — the same approach already used for `linkedin`/`google`, since ogen models distinct REST paths, not one path dispatching on query string.
- Fixtures under `internal/provider/indeed/testdata/` were captured live against the production API (verified via `hurl --test`).

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/provider/indeed/... ./internal/openingsmcp/... ./cmd/openings-mcp/...`
- [x] `hurl --test internal/provider/indeed/testdata/*.hurl` against the live Indeed API
- [x] `cmd/indeed` debug CLI run live (`--keywords "software engineer" --location Taipei --country Taiwan --fetch-details 1`) — real results with full description
- [x] `docker run ... openapi-spec-validator internal/provider/indeed/openapi.yaml` — OK


amikai:
**Note on information loss vs. python-jobspy:** jobspy's Indeed scraper does discard data. Its GraphQL query already asks for `ceoName`/`ceoPhotoUrl`, `headerImageUrl`, employer `addresses`, `recruit.detailedSalary`/`workSchedule`, and `dateOnIndeed`, but `_process_job()` in `jobspy/indeed/__init__.py` never maps them, because jobspy's `JobPost` is one dataclass shared across LinkedIn/Indeed/ZipRecruiter/Glassdoor/etc. and only carries fields that fit that common shape.

`indeed.JobDetail` in this PR has no such constraint — it's Indeed-only. So this PR now requests and surfaces all of the above (plus `source.name`, a structured `Location`) on `indeed_get_job_detail`, rather than inheriting jobspy's field selection. `indeed_search_jobs` (the summary call) intentionally stays lean, matching the search/detail split every other provider in this repo already uses.

Flagging here in case a future pass wants to audit the other providers (particularly ones with a python-jobspy or similar reference implementation) for the same pattern — reference-scraper field selection isn't necessarily the ceiling of what the upstream API actually returns.

internal/provider/indeed/schema.graphql
```diff
@@ -0,0 +1,183 @@
...
+}
+
+type Range {
+  min: Float
+  max: Float
+}
+
+# Live error: "Cannot query field min on type RangeType" when selecting
+# min/max without a fragment — range is RangeType, concrete type is Range.
+union RangeType = Range
```
amikai: **[P1] Handle all RangeType variants**\n\nThe live API returns `AtLeast` and accepts `AtMost`, but this union declares only `Range`. Because genqlient adds `__typename`, encountering a one-sided salary aborts the entire search with `unexpected concrete type ... AtLeast`. Add the missing variants, corresponding fragments/parser handling, and a regression fixture.

Makefile
```diff
@@ -7,6 +7,7 @@ OPENAPI_SPECS := \
 	internal/provider/google/openapi.yaml \
 	internal/provider/greenhouse/openapi.yaml \
 	internal/provider/icims/openapi.yaml \
+	internal/provider/indeed/openapi.yaml \
```
amikai: **[P1] Remove the deleted OpenAPI target**\n\nThis PR replaced `openapi.yaml` with `API.md` but still adds the deleted file as a prerequisite. `make validate-openapi` stops with `No rule to make target internal/provider/indeed/openapi.yaml`, which is the current failing CI check.

internal/provider/indeed/country.go
```diff
@@ -0,0 +1,113 @@
...
+// full name: it's Indeed's own (usually two-letter) subdomain code, e.g.
+// Taiwan's is "tw" (tw.indeed.com), Malaysia's is "malaysia"
+// (malaysia.indeed.com), and the US's is "www" (www.indeed.com) — each
+// taken verbatim from jobspy's Country.indeed_domain_value.
+var countries = map[string]Country{
+	"argentina":            {"ar", "AR"},
+	"australia":            {"au", "AU"},
+	"austria":              {"at", "AT"},
+	"bahrain":              {"bh", "BH"},
+	"bangladesh":           {"bd", "BD"},
```
amikai: **[P2] Reject unsupported Indeed countries locally**\n\nSeveral accepted entries are not valid Indeed sites, including Bangladesh, Bulgaria, Croatia, Cyprus, Estonia, Latvia, Lithuania, Malta, Slovakia, and Slovenia. A live Slovenia request returns `BAD_USER_INPUT: Request country SL does not correspond to a valid Indeed site`; restrict this table to supported sites so these fail as unknown countries before making the request.

internal/openingsmcp/indeed.go
```diff
@@ -0,0 +1,270 @@
...
+			JobTypes:     j.JobTypes,
+			Compensation: indeedCompensationFromHTTP(j.Compensation),
+		})
+	}
+	return out
+}
+
+type indeedDetailInput struct {
+	JobKey  string `json:"job_key" jsonschema:"Opaque Indeed job key (key from indeed_search_jobs results)."`
+	Country string `json:"country,omitempty" jsonschema:"Country used to resolve the job's site domain for its URL; should match the country used in the original search. Defaults to Taiwan."`
```
amikai: **[P2] Do not default detail calls to Taiwan**\n\n`jobData` is country-scoped: a live US key returns a result with `indeed-co: US` but an empty list with `indeed-co: TW`. Since this field is optional, non-Taiwan detail calls that omit it become false not-found errors. Make country required or carry an unambiguous country value in each search result.

internal/provider/indeed/parse.go
```diff
@@ -0,0 +1,204 @@
...
+	cur := currency
+	if bs == nil && estimated != nil {
+		bs = estimated
+		cur = estimatedCurrency
+	}
+	if bs == nil {
+		return nil
+	}
+	comp := &Compensation{Currency: cur, Interval: strings.ToUpper(bs.UnitOfWork)}
+	if bs.Range.Min != 0 {
```
amikai, line 31: **[P2] Preserve fractional compensation**\n\nIndeed returns fractional hourly ranges in production, such as 22.5–25 and 24–27.5. Converting these floats to `int` silently reports 22–25 and 24–27. Keep the amounts as `float64` or another exact numeric representation through the provider and MCP output.

internal/provider/indeed/client.go
```diff
@@ -0,0 +1,212 @@
...
+// Jobs searches jobSearch with r's criteria.
+func (c *Client) Jobs(ctx context.Context, r *JobsRequest) (*JobsResponse, error) {
+	country, err := resolveCountry(r.Country)
+	if err != nil {
+		return nil, err
+	}
+	what := r.Keywords
+	var location *JobSearchLocationInput
+	if r.Location != "" {
+		radius := r.RadiusMiles
```
amikai: **[P2] Respect an explicit zero-mile radius**\n\nIndeed accepts radius 0 and uses it for exact-location searches, but this code silently expands it to 25 miles. Both the MCP schema and debug CLI permit an explicit zero, so callers receive materially broader results than requested. Distinguish omitted radius from zero or disallow zero in the public inputs.

internal/openingsmcp/indeed.go
```diff
@@ -0,0 +1,270 @@
...
+	Location     *indeedLocation     `json:"location,omitempty"`
+	PostedDate   string              `json:"posted_date,omitempty"`
+	Description  string              `json:"description,omitempty" jsonschema:"Full job description as HTML, as Indeed sends it."`
+	JobTypes     []string            `json:"job_types,omitempty"`
+	Compensation *indeedCompensation `json:"compensation,omitempty"`
+
+	Source      string `json:"source,omitempty" jsonschema:"Listing source: usually the employer's own name, or a third-party board's name when Indeed aggregated the posting."`
+	DateIndexed string `json:"date_indexed,omitempty" jsonschema:"When Indeed indexed/last refreshed this posting; can be later than posted_date for reposted or refreshed listings."`
+
+	CompanyWebsite     string   `json:"company_website,omitempty"`
```
amikai, line 213: **[P2] Include the company logo in detail output**\n\nThe provider parses `CompanyLogo`, but the MCP detail type and converter discard it despite claiming to surface every requested `jobData` field. Add a `company_logo` output field and map `d.CompanyLogo`; the current test explicitly codifies the unintended loss.

amikai (commented):
Review findings: two blockers and five correctness issues.

internal/provider/indeed/parse.go
```diff
@@ -0,0 +1,225 @@
...
+	company, companyURL := "", ""
+	if j.Employer != nil {
+		company = j.Employer.Name
+		companyURL = companyURLFromRelative(j.Employer.RelativeCompanyPageUrl, siteBase)
+	}
+	var types []string
+	if len(j.Attributes) > 0 {
+		types = make([]string, 0, len(j.Attributes))
+		for _, a := range j.Attributes {
+			types = append(types, a.Label)
```
amikai: **[P2] Filter attributes before labeling them job types**

Indeed’s `attributes` array includes skills, benefits, credentials, and other metadata—not only employment types. A live result produced entries such as `Microsoft Word`, `HIPAA`, `health insurance`, and `data entry` under `job_types`, bloating MCP output and misrepresenting the posting. Filter to recognized employment-type attributes, or expose the full collection under a generic `attributes` field. Apply the same fix to the detail conversion below.

cmd/indeed/main.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		if job.PostedDate != "" {
+			fmt.Fprintf(w, "Posted: %s\n", job.PostedDate)
+		}
+		if len(job.JobTypes) > 0 {
+			fmt.Fprintf(w, "Job types: %v\n", job.JobTypes)
+		}
+		if job.Compensation != nil {
+			c := job.Compensation
+			fmt.Fprintf(w, "Compensation: %g-%g %s (%s)\n", c.MinAmount, c.MaxAmount, c.Currency, c.Interval)
```
amikai: **[P2] Format one-sided salaries without a zero bound**

The new `AtLeast`/`AtMost` decoding works, but this printer always emits both numeric bounds. A live `AtLeast` salary renders as `20-0 USD (HOUR)`, falsely implying a zero maximum; `AtMost` similarly becomes `0-30`. Render one-sided values as `>= 20` / `<= 30`, exact values once, and true ranges as `min-max`.

amikai (commented):
Second-pass review findings: two remaining output-correctness issues.

PR #175, opened 2026-07-15
feat: add iCIMS ATS provider with 390-company roster

amikai:
## Summary
- Add iCIMS as a multi-company ATS adapter for `search_jobs_by_company` / `get_job_detail_by_company`, backed by public `careers-*.icims.com` portal HTML (server-side `searchKeyword` / `searchLocation`, JSON-LD detail).
- Ship provider package (`internal/provider/icims`), fixture-backed tests, `cmd/icims` debug CLI, registry / verify-companies wiring, and README coverage.
- Curate a 390-company roster, each checked live with `verify-companies --provider icims` (0 errors).

## Notes
- Official Talent Cloud API is partner-gated; this uses the public iframe careers portal only.
- Upstream page size varies by tenant (20 or 50); the adapter stitches into the unified page size of 20.
- `Filters()` is empty (no stable cross-tenant facet JSON API); keyword + location are server-side.
- Uncurated `*.icims.com` careers URLs still resolve via `ParseCareersURL`.

## Test plan
- [x] `go test ./internal/provider/icims/ ./internal/ats/ ./cmd/openings-mcp/ ./cmd/verify-companies/`
- [x] `go run ./cmd/verify-companies --provider icims` → 390 ok, 0 error
- [x] Live CLI smoke: `go run ./cmd/icims --company Peraton search` / `detail --id …`
- [ ] MCP: `search_jobs_by_company` for a roster name (e.g. Peraton) and a raw `careers-*.icims.com` URL


internal/provider/icims/client.go
```diff
@@ -0,0 +1,199 @@
...
+	if req.Page > 0 {
+		q.Set("pr", strconv.Itoa(req.Page))
+	} else {
+		q.Set("pr", "0")
+	}
+	if kw := strings.TrimSpace(req.Keyword); kw != "" {
+		q.Set("searchKeyword", kw)
+	}
+	if loc := strings.TrimSpace(req.Location); loc != "" {
+		q.Set("searchLocation", loc)
```
amikai: **[P1] Resolve location labels before sending `searchLocation`.** iCIMS expects the encoded `<option>` value here (for example `12781-12827-Austin`), not free text such as `Austin`. In a live BusPatrol reproduction, an unfiltered request and `searchLocation=Austin` both returned job IDs `1977,1925,1922` (including the Lorton job), while the encoded value returned only the Austin jobs `1977,1922`. As written, `SearchParams.Location` is silently ignored. Please resolve user text against the location option labels/values first, or apply local filtering, and add a location fixture.

internal/ats/icims.go
```diff
@@ -0,0 +1,256 @@
...
+	res, err := client.Search(ctx, &icims.SearchRequest{
+		Keyword:  baseReq.Keyword,
+		Location: baseReq.Location,
+		Page:     guessPr,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("icims: search %q: %w", host, err)
+	}
+
+	upSize := res.PageSize
```
amikai: **[P1] Do not infer the configured upstream page size from the current page length.** A partial last page (or an out-of-range guessed page) makes `res.PageSize` smaller than the tenant's real page size. With a synthetic 60-job tenant configured for 50 jobs per upstream page, unified page 2 guesses `pr=1`, observes 10 jobs, computes `correctPr=2`, refetches an empty page, and returns zero jobs instead of IDs 21–40. Determine the stable page size from a known full page before translating unified offsets.

internal/ats/icims.go
```diff
@@ -0,0 +1,256 @@
...
+			return nil, fmt.Errorf("icims: search %q page %d: %w", host, correctPr+1, err)
+		}
+		selected = append(selected, more.Jobs...)
+	}
+
+	if len(selected) > pageSize {
+		selected = selected[:pageSize]
+	}
+
+	total := totalPagesUp * upSize
```
amikai: **[P2] This does not produce an exact total when the final upstream page is partial.** For a 60-job tenant with two 50-item upstream pages, this reports `TotalCount=100` and five unified pages instead of 60 and three. Even after fixing page-size discovery, the last page length is needed for an exact total; otherwise the API advertises empty pages. Please fetch the final upstream page when necessary or change the returned metadata so an estimate is not presented as exact.

amikai (commented):
Inline findings from the iCIMS provider review.

internal/provider/icims/parse.go
```diff
@@ -0,0 +1,390 @@
...
+		lv := strings.ToLower(o.Value)
+		ll := strings.ToLower(o.Label)
+		if !strings.Contains(ll, lower) && !strings.Contains(lv, lower) {
+			continue
+		}
+		rank := len(o.Label)
+		if rank == 0 {
+			rank = len(o.Value)
+		}
+		if best == nil || rank < best.rank || (rank == best.rank && o.Value < best.value) {
```
amikai: **[P1] Preserve all fuzzy location matches instead of selecting one arbitrary option.** The unified `Location` contract is a fuzzy match, but this ranking collapses broad inputs such as a country or state to one portal option. On the live BusPatrol board, the unfiltered result has three US jobs (Austin and Lorton), while `Location: "US"` resolves to the Austin option and returns only the two Austin jobs. Similar substring collisions can also misroute short state codes. Please fan out/merge all matching option values or apply filtering in a way that retains every matching location.

internal/ats/icims.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func (a *ICIMSAdapter) Search(ctx context.Context, slug string, p SearchParams) (*SearchResult, error) {
+	host, companyName, err := a.resolveHost(slug)
+	if err != nil {
+		return nil, err
+	}
+
+	page := clampPage(p.Page)
+	start := (page - 1) * pageSize
```
amikai: **[P2] Guard this multiplication against integer overflow.** `Page` is user-controlled and has no maximum in the tool schema. With `page=math.MaxInt`, `start` wraps to `-40`, bypasses the `start >= total` check, and later panics at `res.Jobs[offsetInPage:]` with `slice bounds out of range [-40:]`. Reject pages above `math.MaxInt/pageSize` before multiplying, matching the existing Workday adapter's guard.

amikai (commented):
Second-pass inline findings for the iCIMS provider.

internal/ats/icims.go
```diff
@@ -0,0 +1,394 @@
...
+			} else {
+				return &SearchResult{Jobs: nil, TotalCount: 0, Page: page, TotalPages: 0}, nil
+			}
+		}
+	}
+
+	// Broad fuzzy inputs can hit several portal options. Use each encoded
+	// option as the source of truth because listing-card text may omit country
+	// or state tokens, while the provider client bounds matches and requests.
+	if len(locValues) > 1 {
```
amikai: **[P1] Avoid failing broad location searches with full-board fan-out.** Every multi-option location is expanded into separate full-board crawls, with hard limits of 10 matches and 50 requests. This breaks ordinary fuzzy searches: on the live Peraton board, `Location: "CA"` matches 14 options and returns `location is too broad`. iCIMS accepts repeated `searchLocation` parameters with OR semantics—I verified that a repeated Los Angeles + San Diego request returns their native union—so the adapter can send all resolved values in one paginated query instead of fanning out.

internal/provider/icims/parse.go
```diff
@@ -0,0 +1,420 @@
...
+			continue
+		}
+		parts := make([]string, 0, 3)
+		for _, key := range []string{"addressLocality", "addressRegion", "addressCountry"} {
+			if s := strings.TrimSpace(asString(addr[key])); s != "" && !strings.EqualFold(s, "UNAVAILABLE") {
+				parts = append(parts, s)
+			}
+		}
+		if len(parts) > 0 {
+			return strings.Join(parts, ", ")
```
amikai: **[P2] Preserve every job location.** This returns immediately after formatting the first `jobLocation` entry, silently discarding the rest. Multi-location postings are common: live Peraton job 168635 contains Herndon and Annapolis Junction in JSON-LD, but the adapter returns only Herndon. Please collect and join all non-empty locations, as the other multi-location adapters do, and cover an array with at least two places in the parser test.

internal/ats/icims.go
```diff
@@ -0,0 +1,394 @@
...
+	if strings.HasSuffix(key, ".icims.com") && key != "icims.com" {
+		return key, "", nil
+	}
+	return "", "", fmt.Errorf("icims: unknown company %q; pass a roster career-portal host or a *.icims.com careers URL", slug)
+}
+
+func icimsJobSummaries(jobs []icims.Job, host string) []JobSummary {
+	out := make([]JobSummary, 0, len(jobs))
+	for _, j := range jobs {
+		out = append(out, JobSummary{
```
amikai, line 341: **[P2] Populate posting dates in search summaries.** `icimsJobSummaries` never sets `PostedAt`, even though many iCIMS listing cards expose a posted date in the span following the `Requisition Post Information*: Posted Date` label. The unified contract says to return an ISO date when the upstream provides one, so these searches currently lose available data. Please parse the optional card date into `icims.Job`, normalize it here, and add a fixture-backed assertion.

amikai (commented):
Third-pass review findings for the iCIMS provider.

PR #182, opened 2026-07-16
fix(mcp): improve tool schemas and routing guidance

amikai:
## Context

MCP tool descriptions and schemas guide host models when selecting tools and constructing arguments. Several descriptions had drifted from the actual provider behavior, and the server-level routing instructions contained conflicting guidance for named companies.

## What changed

- correct the 104 remote-filter description so omission means all workplace arrangements, not on-site only
- clarify that unified company URL lookup only accepts recognized careers URLs on supported ATS providers
- disambiguate named-company routing from explicitly selected job boards or careers sites
- update unified company coverage wording from hundreds to thousands of curated companies
- remove the unused `next_cursor` field from unified company search output
- add stable defaults and validation constraints such as `minLength`, `minItems`, and `uniqueItems`, while avoiding redundant constraints on enum fields
- add missing free-text and array constraints to LinkedIn and Indeed search schemas
- describe NVIDIA `external_path` as the identifier for its detail tool
- expand the MCP server contract test to cover Indeed and validate tool schemas, descriptions, output schemas, annotations, and routing guidance

## Root cause

The schemas and server instructions evolved independently from provider and ATS behavior. This left stale assumptions in the LLM-facing contract, including an unsupported arbitrary-careers-URL promise and overlapping routing rules.

## Impact

Host models should select the intended tool family more consistently, avoid unsupported URL calls and empty filter values, and construct pagination and provider-specific detail requests with clearer defaults and identifiers.

## Validation

- `go test ./...`
- `go vet ./...`
- `git diff --check`


internal/openingsmcp/job104.go
```diff
@@ -14,11 +14,13 @@ var job104SearchInputRawSchema = []byte(`{
...
 		"keyword": {
 			"type": "string",
-			"description": "Free-text keyword search."
+			"description": "Free-text keyword search.",
+			"minLength": 1
 		},
 		"area": {
 			"type": "string",
 			"description": "City/region filter.",
+			"minLength": 1,
```
amikai: `area` is enum-constrained, so every legal value already has length ≥ 1 — this `minLength` is dead weight. Same for `country` in nvidia.go and `location` in tsmc.go; the plain free-text fields are the only ones that need it.

internal/openingsmcp/nvidia.go
```diff
@@ -48,6 +49,7 @@ var nvidiaSearchInputRawSchema = []byte(`{
 		"country": {
 			"type": "string",
 			"description": "Country filter.",
+			"minLength": 1,
```
amikai: Same as job104 `area`: the enum already guarantees non-empty, so this `minLength` is redundant.

internal/openingsmcp/tsmc.go
```diff
@@ -18,11 +18,13 @@ var tsmcSearchInputRawSchema = []byte(`{
...
 		"keyword": {
 			"type": "string",
-			"description": "Free-text keyword search across job titles."
+			"description": "Free-text keyword search across job titles.",
+			"minLength": 1
 		},
 		"location": {
 			"type": "string",
 			"description": "Location filter.",
+			"minLength": 1,
```
amikai: Same as job104 `area`: the enum already guarantees non-empty, so this `minLength` is redundant.

amikai (commented):
Two non-blocking notes that have no diff line to anchor on:

- **LinkedIn was skipped in the constraint sweep**: `internal/openingsmcp/linkedin.go` still has `keyword`/`location` without `minLength` and `company_ids` without `minItems`/`uniqueItems`; Indeed's search input similarly lacks `minLength` on `keyword`/`location`/`cursor`. Fine as a follow-up.
- **PR body is stale**: it still says the curated-only URL limitation is documented "for Eightfold and SAP SuccessFactors", but the final commits deliberately removed the provider names (and the tests now assert their absence).

PR #183, opened 2026-07-16
docs(go): improve doc comments

amikai:
## Context

The handwritten Go packages and commands had uneven package documentation, and several search-to-detail APIs did not clearly connect the identifier returned by search with the value required by the corresponding detail call.

## Changes

- add package comments for command and handwritten provider packages
- document the unified ATS search, filter, and detail data flow with Go doc links
- link provider-specific detail keys such as job IDs, Indeed key/country pairs, and Workday external paths
- normalize a few existing comments to Go doc syntax without adding comments that merely restate symbol names

## Impact

`go doc`, editors, and pkgsite-style renderers now expose the important cross-call contracts directly, while implementation methods remain free of redundant comments. Runtime behavior is unchanged.

## Validation

- `go test ./...`
- `go vet ./...`
- `golangci-lint run --enable-only staticcheck --enable-only godot --output.text.path stdout ./...`
- `git diff --check`


internal/ats/ats.go
```diff
@@ -13,27 +13,30 @@ import (
...
+	// Pass a result's [JobSummary.JobID] and the same slug to [Adapter.Detail].
 	Search(ctx context.Context, slug string, p SearchParams) (*SearchResult, error)
 	// Filters returns the filter dimensions and values available for the company
-	// identified by slug.
+	// identified by slug. Pass its keys and values through
+	// [SearchParams.Filters] on a later call to [Adapter.Search].
 	Filters(ctx context.Context, slug string) (FilterSet, error)
-	// Detail returns the full posting for jobID at the company identified by slug.
+	// Detail returns the full posting identified by [JobSummary.JobID].
+	// jobID and slug must come from the same [Adapter.Search] call.
```
amikai: nit: slug is an input to Search, not an output, so "come from the same call" reads loosely. Consider "jobID must come from a Search performed with the same slug".

internal/provider/tsmc/client.go
```diff
@@ -96,6 +96,7 @@ type Client struct {
 	baseURL    string
 }
 
+// JobsRequest treats Page as one-based and PerPage as the requested page size.
```
amikai: nit: sibling comments in this PR document zero behavior (google Page, indeed Limit), but this one omits that PerPage <= 0 falls back to defaultPerPage (10). Worth adding for symmetry.

PR #192, opened 2026-07-16
refactor(ats): parse careers URLs with documented regexes

amikai:
## Summary
- Replace multi-step host/path splitting in `ParseCareersURL` with one-shot regex matchers across adapters (Greenhouse, Lever, Ashby, SmartRecruiters, Workday, Oracle, Eightfold, Recruitee, iCIMS, Teamtailor).
- Document each regex variable with real careers URL examples; SuccessFactors stays roster-only (no shared host pattern) with examples on the method doc.
- Add shared `matchCareersSlug` helper for host+path slug adapters.

## Test plan
- [x] `go test ./internal/ats/ ./internal/provider/workday/ -count=1`
- [ ] Spot-check a few live careers URLs still resolve via MCP company tools

Made with [Cursor](https://cursor.com)


internal/ats/careersurl.go
```diff
@@ -27,6 +28,26 @@ func parseCareersInput(s string) (*url.URL, bool) {
...
+// matchCareersSlug matches re against lowercase(host)+escapedPath and
+// returns the URL-decoded first capture group.
+//
+// Example:
+//
+//	re := regexp.MustCompile(`(?i)^jobs\.ashbyhq\.com/([^/]+)`)
+//	u, _ := url.Parse("https://jobs.ashbyhq.com/Acme%20Inc")
+//	slug, ok := matchCareersSlug(re, u) // "Acme Inc", true
+func matchCareersSlug(re *regexp.Regexp, u *url.URL) (string, bool) {
+	m := re.FindStringSubmatch(strings.ToLower(u.Hostname()) + u.EscapedPath())
```
amikai, line 40: Behavior change hiding in the refactor: `firstPathSegment` skipped empty segments, so `https://jobs.lever.co//acme` used to parse to `acme`. Here `[^/]+` must start immediately after the host's `/`, so double-slash URLs now reject (Workday has the same change — `.com//Site` was accepted via `strings.Trim`). Stricter is defensible, but worth a test pinning whichever behavior is intended.

internal/ats/careersurl.go
```diff
@@ -27,6 +28,26 @@ func parseCareersInput(s string) (*url.URL, bool) {
...
+	slug, err := url.PathUnescape(m[1])
+	if err != nil || slug == "" {
+		return "", false
+	}
+	return slug, true
+}
+
 // firstPathSegment returns the first non-empty path segment, URL-decoded,
 // or "" when the path has none (or decoding fails).
 func firstPathSegment(u *url.URL) string {
```
amikai: This PR removes `firstPathSegment`'s last production callers — only the fake adapter in `registry_test.go` and its own unit test remain. Consider moving it into the test file (or inlining it into `fakeAdapter`) instead of keeping a test-only helper here.

internal/ats/icims.go
```diff
@@ -17,6 +18,16 @@ import (
...
 var _ Adapter = (*ICIMSAdapter)(nil)
 
+// icimsCareersHostRE matches *.icims.com hosts. login*, cdn*, api.*, and
+// www.icims.com are rejected after the match — they are not job boards.
+//
+// Examples (hostname):
+//   - careers-peraton.icims.com
+//   - uscareers-example.icims.com
+var icimsCareersHostRE = regexp.MustCompile(
+	`(?i)^[a-z0-9](?:[a-z0-9.-]*[a-z0-9])?\.icims\.com$`,
```
amikai, line 31: Nit: this is quietly stricter than the old `HasSuffix` check (host must now start/end alphanumeric). Since iCIMS has no roster gate — the host is the slug — this regex is the only validation, so a word in the doc comment that the tightening is intentional would help.

amikai (commented):
Checked the regexes against the old logic — host sets (incl. EU variants), anchoring against host-spoofing, Workday locale handling, and Oracle segment boundaries are all equivalent and covered by the existing tests. Two findings and one nit inline.

Issue #196, opened 2026-07-16
successfactors: job detail fails for tenants using <h1 itemprop="title"> (e.g. jobs.telefonica.com)


amikai:
## Summary

`get_job_detail_by_company` (and the underlying `successfactors.Client.JobDetail`) fails with `unrecognized detail page` for every job on `jobs.telefonica.com`, and likely other tenants with the same template variant.

## Root cause

`parseJobDetailHTML` in [internal/provider/successfactors/parse.go](internal/provider/successfactors/parse.go) only matches the title element via:

```go
title := strings.TrimSpace(doc.Find(`span[itemprop="title"]`).First().Text())
```

`jobs.telefonica.com`'s detail-page template renders the title as:

```html
<h1 id="job-title" itemprop="title">Ejecutivo Asociado de Nómina</h1>
```

— an `<h1>`, not a `<span>` — so the selector never matches, `title` comes back empty, and `parseJobDetailHTML` returns `ok=false`, surfacing as `job detail "<id>": unrecognized detail page`.

## How it was found

Discovered during v0.8.0 release smoke testing (`.claude/skills/release`), sampling 4 random Telefónica postings — all 5 tested job IDs (search + 4 detail follow-ups) failed identically, so it's 100% reproducible for this tenant, not upstream flakiness.

## Why it wasn't caught earlier

`cmd/verify-companies` only exercises `Adapter.Search`, never `Adapter.Detail` (see `cmd/verify-companies/main.go:236`), so a tenant whose detail-page template diverges from the `span[itemprop="title"]` assumption slips through roster verification undetected.

## Suggested fix

Widen the title selector to accept either tag, e.g. `doc.Find(`[itemprop="title"]`).First()` (drop the `span` element constraint), and add a regression test fixture using an `<h1 itemprop="title">` template (mirroring the existing `client_test.go` "unrecognized detail page" case). Worth spot-checking other successfactors roster tenants for the same `h1` vs `span` divergence.

## Scope

Isolated to `internal/provider/successfactors`; unrelated to the current release's actual changes (`git log` on that path shows no recent title-parsing changes). Not release-blocking on its own since search still works for every tenant tested.

PR #198, opened 2026-07-16
feat(verify-companies): probe Detail after each successful search

amikai:
Search-only verification let #196 slip through: jobs.telefonica.com passed the roster sweep while every JobDetail call failed with "unrecognized detail page". Each successful search is now followed by one `Adapter.Detail` probe on the first job of page 1 (zero-job boards skipped), reported as a new `DETAIL_ERROR` status that keeps the job count, carries the probed job ID plus error, and fails the exit code. Each of the two requests gets its own `--timeout`.

Live successfactors sweep confirms the probe catches Telefónica — and shows #196 is a common template variant, not a one-off: 14 of 76 tenants fail the detail probe with the same error (Bunge, Deere, Endress+Hauser, Gulfstream, Knorr-Bremse, Lincoln Financial, MediaMarktSaturn, Mondi, NextEra Energy, Nucor, OMV, PACCAR, Super Micro, Telefónica), with zero search errors.


cmd/verify-companies/main.go
```diff
@@ -226,32 +232,58 @@ func runChecks(ctx context.Context, checks []check, timeout time.Duration, concu
...
-	res, err := c.adapter.Search(ctx, c.slug, ats.SearchParams{Page: 1})
+	res, err := c.search(ctx, timeout)
 	if err != nil {
 		r.Status, r.Detail = statusError, err.Error()
 		return r
 	}
 	r.Status = statusOK
 	r.Jobs = res.TotalCount
+
+	if len(res.Jobs) == 0 {
```
amikai: `len(res.Jobs) == 0` does not necessarily mean this is a zero-job board. Some adapters can report `TotalCount > 0` while filtering out page-one postings that lack a usable `JobID` (for example, Workday skips entries without `externalPath`). In that case this returns `OK` and bypasses the Detail probe this PR is intended to enforce. Please only treat `TotalCount == 0` as zero-job; if the total is nonzero but no usable summary is available, report an error or otherwise obtain a probeable job.

PR #199, opened 2026-07-16
fix: roster prune + clear verify DETAIL_ERRORs

amikai:
## Summary

Follow-up from a full `verify-companies` sweep across all roster providers.

- **Ashby roster:** remove 2 boards confirmed 404 (`Continue`, `Roo Code`). No supported-ATS migration found for those slugs.
- **Eightfold:** `publicUrl` may be JSON `null` (NetApp, Houston ISD); make it nullable and fall back to `origin + positionUrl`.
- **Workday:** accept location-less `externalPath` values (`/job/{titleSlug}`) used by tenants such as BSI, LVHN, VUMC.
- **SuccessFactors:** multi-brand tenants omit `itemprop="title"`; fall back to `og:title` / document title (Endress+Hauser, Knorr-Bremse, MediaMarktSaturn).
- **iCIMS:** some portals serve live iframe detail without JobPosting JSON-LD; scrape `h1.iCIMS_Header` + expandable text.
- **verify-companies:** auto-cap Recruitee at 2 concurrent workers so default concurrency does not trip HTTP 429s.

## Test plan

- [x] `go test` for affected packages (workday, successfactors, icims, eightfold, ats, verify-companies)
- [x] `go run ./cmd/verify-companies --provider ashby` → 1275 OK, 0 error
- [x] eightfold, successfactors, icims, workday, recruitee re-sweeps clean (or only transient 5xx cleared on retry)


cmd/verify-companies/main.go
```diff
@@ -216,15 +216,24 @@ func buildChecks(adapters []ats.Adapter) []check {
...
 	sem := make(chan struct{}, concurrency)
+	recruiteeCap := min(2, concurrency)
+	recruiteeSem := make(chan struct{}, recruiteeCap)
 	var wg sync.WaitGroup
 	for i, c := range checks {
 		wg.Go(func() {
 			sem <- struct{}{}
 			defer func() { <-sem }()
+			if c.adapter.Name() == "recruitee" {
+				recruiteeSem <- struct{}{}
```
amikai, line 235: [P2] Acquire provider cap before global worker slot\n\nA Recruitee goroutine acquires the global semaphore before waiting on the two-worker Recruitee semaphore. With enough Recruitee checks, idle waiters can occupy every global slot and prevent SuccessFactors, Teamtailor, and Workday checks from starting until the 62-entry Recruitee roster drains. Acquire the Recruitee cap first, then the global slot, or use separate bounded pools.

cmd/eightfold/main.go
```diff
@@ -445,8 +445,11 @@ func printDetail(d eightfold.PositionDetail, format string) error {
 		fmt.Printf("Department: %s\n", d.Department.Value)
 	}
 	fmt.Printf("Posted: %s\n", time.Unix(int64(d.PostedTs), 0).UTC().Format("2006-01-02"))
-	if d.PublicUrl != "" {
-		fmt.Printf("URL: %s\n", d.PublicUrl)
+	if u, ok := d.PublicUrl.Get(); ok && u != "" {
+		fmt.Printf("URL: %s\n", u)
+	} else if d.PositionUrl != "" {
+		// Some tenants send publicUrl: null; positionUrl is site-relative.
+		fmt.Printf("URL: %s\n", d.PositionUrl)
```
amikai: [P2] Keep fallback URL absolute\n\nWhen publicUrl is null—the NetApp/Houston case this patch addresses—positionUrl is site-relative, so text output becomes URL: /careers/job/.... The detail command advertises a public URL and search results already emit absolute links. Pass the tenant base URL into printDetail or compose the fallback before printing.

internal/provider/workday/openapi.yaml
```diff
@@ -136,8 +136,11 @@ paths:
       summary: Get job detail
       description: >
         Fetches a single job posting. `externalPath` from the search
-        response has the form `/job/{location}/{titleSlug}`, with exactly
-        two segments after `/job/`. Split it on the first `/` and pass the
+        response is usually `/job/{location}/{titleSlug}` (two segments
+        after `/job/`). Some tenants omit the location segment and emit
+        `/job/{titleSlug}` only — pass an empty `location` in that case
+        (the client encodes it as `/job//{titleSlug}`, which the CXS API
```
amikai, line 142: [P3] Regenerate Workday client artifacts\n\nThe OpenAPI description now documents location-less paths, but the checked-in generated client still says externalPath always has exactly two segments. Running go generate ./internal/provider/workday modifies five oas_*_gen.go files; commit that output so generated documentation remains synchronized.

PR #201, opened 2026-07-17
feat: add Jobindex.dk job board provider

amikai:
## Summary

- Add a dedicated **Jobindex** (jobindex.dk) job-board provider with `jobindex_search_jobs` / `jobindex_get_job_detail` MCP tools, wired into the openings-mcp server.
- Search parses the public `/jobsoegning` HTML page’s embedded `var Stash` payload (the old `.json` endpoint returns 204); detail uses `/vis-job/{tid}` so requests stay on Jobindex instead of following off-site `/jobannonce/` redirects.
- Ship a `cmd/jobindex` debug CLI (`search` / `detail`), fixture-backed unit/MCP tests, hurl captures, and README / server-instruction updates. Implementation referenced the [ai-job-search](https://github.com/MadsLorentzen/ai-job-search) `jobindex-search` skill.

## Test plan

- [x] `go test ./internal/provider/jobindex ./internal/openingsmcp ./cmd/openings-mcp`
- [x] Live CLI: `go run ./cmd/jobindex search --keyword backend --jobage 14`
- [x] Live CLI: `go run ./cmd/jobindex detail --tid h1683131`
- [x] Live MCP smoke: `jobindex_search_jobs` + `jobindex_get_job_detail` via stdio
- [ ] CI green on this PR


internal/provider/jobindex/client.go
```diff
@@ -0,0 +1,211 @@
...
+	Page int `json:"page"`
+}
+
+// JobDetail is scraped from the /vis-job/{tid} HTML page. Jobindex does not
+// expose a JSON detail API for that view; field names mirror the Stash search
+// result keys where the same concept exists (tid, headline, area, firstdate,
+// share_url, apply_url, company). No fields are synthesized by merging
+// deadlines (e.g. we never invent "ASAP").
+type JobDetail struct {
+	Tid        string         `json:"tid"`
```
amikai: This struct's field alignment is stale — `gofmt -l` flags the file (the padding here is one column too wide). Will likely fail the CI format check; `gofmt -w` fixes it.

internal/provider/jobindex/parse.go
```diff
@@ -0,0 +1,270 @@
...
+			}
+			out[k] = v
+		}
+		results = append(results, out)
+	}
+	if pageNum < 1 {
+		pageNum = 1
+	}
+	if totalPages == 0 && hitcount > 0 {
+		totalPages = (hitcount + DefaultPageSize - 1) / DefaultPageSize
```
amikai: This synthesizes `total_pages` under the upstream field name, while openapi.yaml says it "is also present in Stash" and the struct comment says field names match the embedded JSON — a client can't tell derived from upstream. Either drop the fallback (let `omitempty` omit it) or document the derivation on `SearchResponse.TotalPages`.

Also: both search fixtures omit `total_pages`, so only this fallback path is tested — the pass-through path never is.

internal/provider/jobindex/parse.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	if a := doc.Find("a.seejobdesktop, a.seejobmobil").First(); a.Length() > 0 {
+		if href, ok := a.Attr("href"); ok {
+			d.ApplyURL = href
+		}
+	}
+
+	var paras []string
+	doc.Find("div.PaidJob p, div.jix_robotjob-inner p").Each(func(_ int, s *goquery.Selection) {
```
amikai, line 213: The `jix_robotjob` selectors (here and `--area` above) are untested: the only detail fixture is a PaidJob (`h1683131`). A small `r*`-tid fixture would cover the aggregated-posting branch.

internal/provider/jobindex/parse.go
```diff
@@ -0,0 +1,270 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/PuerkitoBio/goquery"
+	"golang.org/x/net/html"
+)
+
+var (
+	reStashMarker = []byte("var Stash = ")
+	reMetaContent = regexp.MustCompile(`(?i)<meta[^>]+content="([^"]*)"[^>]+property="([^"]+)"|<meta[^>]+property="([^"]+)"[^>]+content="([^"]*)"`)
```
amikai: Detail parsing already builds a goquery document — `doc.Find("meta[property]")` + `Attr` would replace this regex, which only handles double-quoted attributes in two orderings.

Nit: `reStashMarker` above has an `re` prefix but is a `[]byte`, always used as `string(...)` — a plain string const would do.

internal/provider/jobindex/client.go
```diff
@@ -0,0 +1,211 @@
...
+func (c *Client) searchURL(req *JobsRequest) (string, error) {
+	u, err := url.Parse(c.baseURL)
+	if err != nil {
+		return "", fmt.Errorf("parse base URL: %w", err)
+	}
+	path := searchPath
+	if area := strings.Trim(req.Area, "/"); area != "" {
+		path = searchPath + "/" + area
+	}
+	u.Path = path
```
amikai: `u.Path = path` drops any path prefix in the base URL; siblings use `u.JoinPath`. Only matters for a proxied `--base-url`, but it's a convention divergence.

internal/openingsmcp/jobindex.go
```diff
@@ -0,0 +1,164 @@
...
+		Keyword:    in.Keyword,
+		Area:       in.Area,
+		Page:       in.Page,
+		JobAgeDays: in.JobAgeDays,
+		Sort:       in.Sort,
+	}
+	if req.Sort == "" {
+		req.Sort = jobindex.SortScore
+	}
+	if req.Sort != jobindex.SortScore && req.Sort != jobindex.SortDate {
```
amikai: Sort is validated three times: the schema `enum`, here, and the client's `searchURL`. This one can go — the client's error already propagates through `errorResult`.

amikai (commented):
Provider layering, schema conventions, and test structure all match the sibling providers; tests pass at all three layers and `go vet` is clean. Inline comments below.

One non-code nit: the test plan says `detail --id h1683131`, but the actual flag is `--tid`.

PR #207, opened 2026-07-17
feat: integrate UltiPro (UKG Pro) ATS provider

amikai:
## Summary
- Adds UltiPro (UKG Pro Recruiting) as a new multi-tenant ATS behind `search_jobs_by_company` / `get_filters_by_company` / `get_job_detail_by_company`
- No usable public API exists (UKG's developer API is OAuth-gated per-tenant), so search/filters are reverse-engineered JSON POST endpoints and job detail is hand-parsed from a JSON literal embedded in server-rendered HTML — the same approach already used for iCIMS and SuccessFactors, so `openapi.yaml` is documentation-only (no `ogen` codegen)
- Seed roster: TechnoServe, Idaho Central Credit Union, Macmillan, SoftWriters, The Salvation Army — each verified live, spanning both observed hosts (`recruiting.ultipro.com` / `recruiting2.ultipro.com`)

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` all pass
- [x] `hurl --test` replays all 8 captured fixtures against the live API
- [x] `make validate-openapi` passes for the new spec
- [x] Live smoke test through the real MCP server (stdio): `search_jobs_by_company` returns real listings for all 5 roster companies, `get_job_detail_by_company` and `get_filters_by_company` verified on live data


internal/ats/ultipro.go
```diff
@@ -0,0 +1,277 @@
...
+	var out []ultipro.SearchFilter
+	for key, values := range filters {
+		switch key {
+		case "department":
+			for _, v := range values {
+				id, err := resolveUltiProCatalogValue(ctx, client.Categories, v, "department")
+				if err != nil {
+					return nil, err
+				}
+				out = append(out, ultipro.SearchFilter{FieldName: 5, Values: []string{id}})
```
amikai: **[P2] Combine same-key filter values into one filter object.** Each selected value currently becomes a separate `TermsSearchFilterDto`, but UltiPro ANDs repeated filters for the same field. Live verification returned Finance=3, IT=1, both IDs in one `Values` array=4, but two separate field-5 objects=0. Resolve all values first and emit one filter per dimension; the same issue applies to `location_type` below.

internal/ats/ultipro.go
```diff
@@ -0,0 +1,277 @@
...
+	filters, err := a.buildFilters(ctx, client, p.Filters)
+	if err != nil {
+		return nil, err
+	}
+	if loc := strings.TrimSpace(p.Location); loc != "" {
+		id, err := resolveUltiProCatalogValue(ctx, client.Locations, loc, "location")
+		if err != nil {
+			return nil, err
+		}
+		filters = append(filters, ultipro.SearchFilter{FieldName: 4, Values: []string{id}})
```
amikai: **[P2] Honor fuzzy and remote location semantics.** The unified tool documents fuzzy location matching and says `remote` matches remote-friendly jobs, but this accepts only one exact physical-location label. Live TechnoServe checks showed the physical `Remote` location returned one job while the field-37 Remote type returned two different remote jobs. Match all catalog labels containing the location into one `Values` array and handle `remote` through the location-type behavior.

internal/ats/ultipro.go
```diff
@@ -0,0 +1,277 @@
...
+// identical to name-based resolution; unknown boards get the canonical URL
+// as a self-describing slug (host, company code, and board id all matter,
+// which a bare code can't carry).
+func (a *UltiProAdapter) ParseCareersURL(u *url.URL) (string, bool) {
+	site, ok := ultipro.ParseCareersURL(u)
+	if !ok {
+		return "", false
+	}
+	if _, ok := ultipro.CompaniesByCode[strings.ToLower(site.CompanyCode)]; ok {
+		return strings.ToLower(site.CompanyCode), true
```
amikai, line 72: **[P2] Preserve alternate board identity.** Any URL whose company code exists in the roster is collapsed to that roster slug even when its host or board ID differs. UltiPro addresses boards using all three fields, so a tenant's second board URL will silently search the curated board instead. Fold only when host, company code, and board ID all match; otherwise retain the canonical URL slug.

internal/ats/ultipro.go
```diff
@@ -0,0 +1,277 @@
...
+		loc := ""
+		if len(o.Locations) > 0 {
+			loc = o.Locations[0].Display()
+		}
+		jobs = append(jobs, JobSummary{
+			JobID:    o.ID,
+			Title:    o.Title,
+			Location: loc,
+			PostedAt: ultiproPostedAt(o.PostedDate),
+			URL:      "", // the search response carries no per-posting URL; Detail's URL is authoritative.
```
amikai: **[P2] Populate search-result URLs.** Every UltiPro search result currently omits its public URL, despite `JobSummary.URL` representing the clickable posting URL and the URL being derivable from the resolved site plus opportunity ID. Pass the site into the summary conversion and construct the same canonical detail URL used by `Detail`.

internal/ats/ultipro.go
```diff
@@ -0,0 +1,277 @@
...
+func (a *UltiProAdapter) Detail(ctx context.Context, slug, jobID string) (*JobDetail, error) {
+	name, site, err := a.resolveSlug(slug)
+	if err != nil {
+		return nil, err
+	}
+	client := ultipro.NewClient(a.baseURL(site), a.hc)
+
+	d, err := client.Detail(ctx, jobID)
+	if err != nil {
+		return nil, fmt.Errorf("ultipro: job %q not found for company %q; pass a job_id exactly as returned by the job search: %w", jobID, slug, err)
```
amikai: **[P2] Do not label every detail failure as not found.** Network errors, timeouts, HTTP 5xx responses, malformed HTML, and `ErrCompanyNotFound` are all reported as an invalid job ID. Only `errors.Is(err, ultipro.ErrJobNotFound)` should produce the teaching error; propagate other failures as fetch errors so outages remain diagnosable.

amikai (commented):
Review findings are attached inline. The full test suite and vet pass, but these behavioral issues should be addressed before merge.

internal/ats/ultipro.go
```diff
@@ -0,0 +1,371 @@
...
+	// already excludes remote makes the combination impossible, so
+	// short-circuit rather than round-trip a query guaranteed to be empty.
+	location := strings.TrimSpace(p.Location)
+	filterInput := p.Filters
+	if strings.EqualFold(location, "remote") {
+		lt := filterInput["location_type"]
+		switch {
+		case len(lt) > 0 && !ultiproContainsFold(lt, "remote"):
+			return &SearchResult{Page: page}, nil
+		case !ultiproContainsFold(lt, "remote"):
```
amikai: **[P2] Intersect `remote` with explicit location-type filters.** When `location="remote"` is combined with `location_type=["Remote", "Hybrid"]`, this branch leaves both values intact and later sends codes `[2, 0]`. Since values within a filter are OR'd, the search returns Hybrid jobs too—live TechnoServe results increased from 2 Remote jobs to 10 Remote-or-Hybrid jobs. These inputs are separate criteria and should intersect to Remote only; mirror the SmartRecruiters adapter by replacing the filter with only Remote when Remote is present, or returning empty when it is absent.

internal/provider/ultipro/client.go
```diff
@@ -0,0 +1,335 @@
...
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, fmt.Errorf("detail: %w", ErrCompanyNotFound)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("detail: HTTP %d", resp.StatusCode)
+	}
+
+	detail, ok := extractOpportunityDetail(body)
+	if !ok {
+		return nil, fmt.Errorf("detail: %w", ErrJobNotFound)
```
amikai: **[P2] Keep parse failures distinct from not-found.** `extractOpportunityDetail` returns false both when the marker is absent—the genuine not-found app shell—and when braces are unbalanced, JSON decoding fails, or required fields are missing. This converts every malformed upstream response into `ErrJobNotFound`, so the adapter's new error branch still reports a bad job ID for parser/upstream regressions. Return a distinct parse error once the marker is present and reserve `ErrJobNotFound` for the absent-marker case.

amikai (commented):
Second-pass review findings are attached inline. Three earlier findings are fully addressed; two edge cases remain.

internal/ats/ultipro.go
```diff
@@ -0,0 +1,372 @@
...
+	// impossible (short-circuit rather than round-trip a query guaranteed to
+	// be empty), and one that includes remote alongside other values (e.g.
+	// ["Remote", "Hybrid"]) must narrow to remote only — verified live,
+	// leaving Hybrid in place turned 2 Remote jobs into 10 Remote-or-Hybrid
+	// jobs.
+	location := strings.TrimSpace(p.Location)
+	filterInput := p.Filters
+	if strings.EqualFold(location, "remote") {
+		if lt := filterInput["location_type"]; len(lt) > 0 && !ultiproContainsFold(lt, "remote") {
+			return &SearchResult{Page: page}, nil
```
amikai, line 131: **[P2] Validate filters before the remote conflict short-circuit.** When `location="remote"` conflicts with `location_type`, this returns an empty result before `buildFilters` validates the supplied filters. Consequently, an invalid location type or an unknown filter key is silently accepted instead of returning the documented teaching error. Validate all supplied filters first, then short-circuit only when the validated location type genuinely excludes Remote.

amikai (commented):
One remaining issue after re-review.

PR #208, opened 2026-07-17
feat: add JOIN (join.com) ATS provider

amikai:
## Summary

- Adds join.com as a new ATS provider, following the standard `integrate-new-provider` pipeline (spec hunt → fixtures → client → provider package → debug CLI → MCP surface).
- join.com's public candidate API (`candidate-api/graphql`) is GraphQL-only, has no server-side keyword search, and never populates job descriptions — so this uses a genqlient client for search (`GetCompanyJobs`, mirroring `internal/provider/indeed`'s genqlient pattern) plus an SSR `__NEXT_DATA__` scrape for job/company detail (mirroring `internal/provider/linkedin`'s HTML-scrape pattern). Full notes in `internal/provider/join/API.md`.
- Wired as a dump-style `ats.Adapter` (like Greenhouse/Lever) since there's no upstream keyword filter — `searchDump` handles query/location/filter matching client-side.
- Seed roster of 5 companies (`internal/provider/join/companies.yaml`), each confirmed live: JOIN, Routine Labs, AnyTax, PALTRON GmbH, hey contact heroes GmbH.
- Registered in the MCP server (`cmd/openings-mcp/main.go`), the careers-URL registry (`internal/ats/registry.go`), and `cmd/verify-companies`.
- New `cmd/join` debug CLI (`companies`/`search`/`get`).

Along the way, live testing caught a real upstream quirk: `PaginationPageInput.page`/`pageSize` are `Float`-typed server-side despite only ever holding integers — an inlined literal coerces silently, but a properly typed `Int!` GraphQL variable gets rejected. Fixed via a `Float → int` genqlient binding so the Go client still works with plain ints.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./...` all pass
- [x] `make hurl-test` — all join.com fixtures replay against live traffic
- [x] Live smoke test through the real MCP server (stdio): `search_jobs_by_company`, `get_job_detail_by_company`, and `get_filters_by_company` against all 5 roster companies, plus careers-URL resolution


amikai (commented):
Two actionable findings from the review:

internal/provider/smartrecruiters/companies.yaml
```diff
@@ -1,21 +1,31 @@
 - company: A Good Company
   company_identifier: AGoodCompany
+- company: A Job Thing
```
amikai: **[P1] Remove the unrelated SmartRecruiters roster commit.** This JOIN provider PR also contains commit `ef4497b`, which adds 100 SmartRecruiters companies. That changes a second `companies.yaml` roster, violates the repository’s single-roster PR convention, and would merge unrelated work. Please rebase the JOIN commit onto `main` or cherry-pick `2dd6d08` onto a clean branch.

internal/ats/join.go
```diff
@@ -0,0 +1,166 @@
...
+			fields["category"] = []string{j.Category}
+		}
+		if j.EmploymentType != "" {
+			fields["employment_type"] = []string{j.EmploymentType}
+		}
+		out = append(out, dumpJob{
+			summary: JobSummary{
+				JobID:    j.IdParam,
+				Title:    j.Title,
+				Location: j.City,
```
amikai: **[P2] Preserve country and remote location metadata.** Reducing the location to `j.City` drops `Country`, `WorkplaceType`, and `RemoteType`. Current roster data includes `REMOTE/ANYWHERE` and `REMOTE/COUNTRY` jobs, so an anywhere-remote role can be displayed as “Berlin,” while searching for “Germany” returns no matches. Please build the displayed/searchable location from city, country, and remote scope, and apply the same mapping to `Detail`.

internal/provider/smartrecruiters/companies.yaml
amikai: Fixed: rebased the JOIN commit onto `main`, dropping `ef4497b`. The branch now has just the two JOIN commits — confirmed no `smartrecruiters` files in the diff.

internal/ats/join.go
amikai: Fixed in 2ce0846. Added `joinLocation()` in internal/ats/join.go, which builds the displayed/searchable location from city + country + remote scope instead of bare city, applied to both Search and Detail. Captured a live REMOTE/ANYWHERE fixture (hey-contact-heroes) to cover it — confirmed live through the real MCP server: `"location":"Remote (Anywhere) · Hamburg, Germany"`. Also found remoteType is present on the job detail SSR page too (not just the search list), so JobDetail now carries it.

PR #214, opened 2026-07-18
feat: integrate Workable as a new ATS provider

amikai:
## Summary

- Add a Workable ATS adapter using the public `apply.workable.com` job-board API (`POST /api/v3/.../jobs` search, filters facets, `GET /api/v2/.../jobs/{shortcode}` detail).
- Ship ogen-generated client, fixture-backed provider tests, seed roster (Blueground, PEOPLECERT, Persado, Skroutz, Zego), and `cmd/workable` debug CLI.
- Wire the adapter into the MCP server registry, `verify-companies`, README provider list, and OpenAPI Makefile targets.

## Design notes

- Multi-company ATS surface via `search_jobs_by_company` (not dedicated tools).
- Official token-gated SPI API was rejected; careers-page public v3/v2 endpoints are used instead.
- Location fields are nullable (`region`/`city` often null on live accounts); OpenAPI uses `OptNilString`.
- Unified page size 20 is served by walking Workable’s fixed 10-item cursor pages.

## Test plan

- [x] `go test ./...`
- [x] `make validate-openapi`
- [x] Live MCP stdio smoke: search all 5 roster companies + Blueground detail + Zego location filter
- [x] Live `cmd/workable` search for persado (null `region`) and zego `--city London`


internal/ats/workable.go
```diff
@@ -0,0 +1,478 @@
...
+		out = append(out, lf)
+	}
+	return out
+}
+
+func workableSummary(account string, j workable.JobSummary) JobSummary {
+	return JobSummary{
+		JobID:    j.Shortcode,
+		Title:    j.Title,
+		Location: workableLocationText(j.Location),
```
amikai, line 345: **[P2] Include secondary job locations**

The summary formats only `j.Location` and ignores `j.Locations`. A live PeopleCert search filtered to London returns jobs whose output location says Tokyo or Stuttgart because London is secondary; `Detail` has the same omission. Combine distinct visible primary and secondary locations.

internal/ats/workable.go
```diff
@@ -0,0 +1,478 @@
...
+		// The total also terminates the walk, so a cursor that stops
+		// advancing cannot loop forever.
+		if !ok || len(items) >= rsp.Total {
+			break
+		}
+		req.Token = workable.NewOptString(token)
+	}
+	jobs := make([]dumpJob, 0, len(items))
+	for _, j := range items {
+		jobs = append(jobs, workableDumpJob(slug, j))
```
amikai, line 203: **[P1] Preserve body-only keyword matches**

Workable’s query searches posting bodies, but `workableDumpJob` leaves `description` empty before `searchDump` reapplies the query. Live `query=engineer` returns 9 upstream matches while MCP returns only 4, dropping valid body-only results. Populate description text—likely through detail requests—before residual filtering, or avoid discarding upstream full-text matches.

PR #219, opened 2026-07-18
feat(jobicy): add provider package and debug CLI

amikai:
Adds Jobicy (remote-jobs board) through the debug-CLI stage: traffic-captured OpenAPI spec + ogen client, fixture-replaying tests, hurl fixtures, and `cmd/jobicy`. MCP wiring is deliberately deferred to a follow-up (handoff notes in `docs/superpowers/plans/2026-07-19-jobicy-provider.md`).

Reviewer notes:

- Spec is written from captured traffic, not the vendor README, which drifts (`id` is a number, `jobIndustry`/`jobType` are string arrays, salary fields are per-row optional).
- One endpoint query-routes jobs vs taxonomies, modeled as a `oneOf` 200 → ogen field-discriminated sum type.
- Zero matches is HTTP 404 with the normal envelope (decoded as a result, not an error); bad filter slugs are 400.
- No detail endpoint upstream: every row carries the full HTML description, so the CLI has no `detail` subcommand.


cmd/jobicy/doc.go
```diff
@@ -0,0 +1,10 @@
+// Command jobicy is a debug CLI for the Jobicy remote-jobs feed.
```
amikai: Convention: `cmd/<name>` gets no doc.go — the `// Command jobicy ...` comment sits directly atop `package main` in main.go (see cmd/apple/main.go). Move this comment there and drop doc.go.

amikai: Fixed in 6e70005 — doc.go removed, comment moved atop main.go.

PR #221, opened 2026-07-18
feat(remotive): add Remotive provider package and debug CLI

amikai:
Adds `internal/provider/remotive` (OpenAPI spec, ogen client, fixture-replaying tests) and `cmd/remotive`, stopping before the MCP surface on purpose (deferred to a follow-up).

The load-bearing finding: Remotive's documented query params (`search`, `category`, `company_name`, `limit`) are all no-ops live — responses come from a Cloudflare cache whose key ignores the query string (byte-identical bodies for every probe, `cf-cache-status: HIT`). So the client models the endpoint as a full dump and filtering is client-side in `remotive.FilterJobs`, which the future `remotive_search_jobs`/`remotive_get_job_detail` tools will reuse. `testdata/jobs_filtered_req.hurl` live-asserts the quirk so `make hurl-test` flags if upstream filtering ever starts working.

Verified: `make validate-openapi`, `go test ./...`, `hurl --test` on the new fixtures, and a live CLI smoke of `search`/`detail`/`categories`.


internal/provider/remotive/mocksrv.go
```diff
@@ -0,0 +1,36 @@
...
+	mux.HandleFunc("/remote-jobs/categories", serveMockJSON(mockCategoriesRsp))
+	return httptest.NewServer(mux)
+}
+
+func serveMockJSON(data []byte) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// Matches the live API's header; the generated decoder strips the
+		// charset parameter before matching.
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+		w.Write(data)
```
amikai, line 34: Bare `w.Write(data)` drops the error, unlike the sibling mocks — amazon's serveMockJSON checks it and apple's uses `_, _ = w.Write(data)`. Consider matching one of them for consistency.

PR #224, opened 2026-07-18
feat(weworkremotely): add provider package and debug CLI

amikai:
## Summary

- Adds `internal/provider/weworkremotely`: a hand-written client for We Work Remotely's public RSS feeds (10 fixed category feeds, each a full unfiltered dump — no ogen/OpenAPI, since RSS/XML isn't a shape ogen generates from).
- Adds `cmd/weworkremotely`: debug CLI (`search`, `detail`, `categories`), smoke-tested live against the real site.
- `FilterJobs` does client-side keyword/category/company/type/region filtering; `Search` fetches only the matching feed when a recognized category is given, otherwise merges all 10.
- `Detail` resolves an id from a fresh full-dump fetch instead of the individual job page — that page sits behind a Cloudflare bot challenge and its embedded JobPosting JSON-LD isn't valid JSON (raw unescaped control characters).
- WWR's documented JSON API (`weworkremotely.com/api`) is write-only for employers (partner-token-gated job posting), confirmed via WebSearch — no public listing/search route exists, so RSS is the only usable read surface, not a shortcut around a better one.

**MCP surface intentionally not wired** (requested: 先不要 wire mcp). A follow-up needs `internal/openingsmcp/weworkremotely.go` + registration in `newServer` + README update before this provider is reachable through the MCP server.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/provider/weworkremotely/... ./cmd/weworkremotely/...`
- [x] `go test ./internal/provider/weworkremotely/...` (mocksrv-backed, all pass)
- [x] `gofmt -l` clean
- [x] `hurl --test` against the 4 fixtures — all pass live against weworkremotely.com
- [x] CLI `search`/`detail`/`categories` smoke-tested against the live site


internal/provider/weworkremotely/client.go
```diff
@@ -0,0 +1,271 @@
...
+// instead of the full 10-feed dump.
+func (c *Client) Search(ctx context.Context, opts FilterOptions) ([]Job, error) {
+	if cat, ok := lookupCategory(opts.Category); ok {
+		jobs, err := c.Jobs(ctx, cat)
+		if err != nil {
+			return nil, err
+		}
+		narrowed := opts
+		narrowed.Category = "" // the fetch already narrowed it; avoid re-filtering
+		return FilterJobs(jobs, narrowed), nil
```
amikai: Dedup behavior depends on the path taken: this recognized-category branch returns the raw `Jobs` feed (which per the docs can list the same job twice), while the full-dump fallback goes through `AllJobs` and dedups by ID. `Search(Category: "Full-Stack Programming")` returns the known duplicate twice, but the same job via the fallback appears once — worth deduping here too for consistency.

internal/provider/weworkremotely/client.go
```diff
@@ -0,0 +1,271 @@
...
+
+// AllJobs fetches every category feed and merges them, deduplicating by
+// [Job.ID]. Categories are independent feeds fetched one at a time, so
+// this is 10 HTTP requests, not 1.
+func (c *Client) AllJobs(ctx context.Context) ([]Job, error) {
+	seen := make(map[string]bool)
+	var all []Job
+	for _, cat := range Categories {
+		jobs, err := c.Jobs(ctx, cat)
+		if err != nil {
```
amikai, line 98: **Correctness/resilience:** `AllJobs` aborts the entire merge on the first per-category fetch error, discarding results from every category that already succeeded. Since `doc.go` documents that this same upstream can intermittently answer some request shapes with a Cloudflare bot-management challenge, a transient hiccup on 1 of these 10 independent feeds turns a no-category `Search` or any `Detail` call into a total failure instead of a degraded-but-usable result. Worth collecting per-category errors and returning the jobs that did succeed (or at least a partial result + error), rather than `return nil, err` on the first failure.

amikai: Fixed in 13cd1f9: `AllJobs` now continues past a failing category and joins every failure into the returned error instead of returning `nil, err` on the first one; `Search`/`Detail` propagate whatever succeeded alongside the joined error. Added 4 tests exercising the partial-failure path (a mock server with one category feed forced to 500).

internal/provider/weworkremotely/client.go
amikai: Fixed in fec1c54: extracted a shared `dedupeByID` helper and applied it on both paths — `AllJobs` now dedupes the merged full dump via the same helper, and `Search`'s recognized-category fast path dedupes the raw `Jobs` result before filtering, so a category-scoped Search no longer returns the known duplicate twice. Added a regression test (`TestClient_Search_categoryPathDedupesLikeFullDump`) against the Full-Stack fixture's real duplicate.

amikai: Fixed in fec1c54 — dedupeByID now applies to the recognized-category fast path too, with a fixture-backed test pinning the 42→41 dedup.

internal/provider/weworkremotely/client.go
amikai: Fixed in 13cd1f9 — AllJobs now accumulates per-category failures with errors.Join and returns whatever succeeded; Search/Detail/CLI propagate the partial-failure semantics, all four paths test-covered.

PR #226, opened 2026-07-18
feat(provider): add Working Nomads client and debug CLI

amikai:
## Summary

- Adds a hand-written client for Working Nomads's public jobs feed (`internal/provider/workingnomads`) plus a debug CLI (`cmd/workingnomads`), following the [integrate-new-provider](.claude/skills/integrate-new-provider/SKILL.md) pipeline through step 5 (debug CLI).
- Recon (documented in `doc.go`) found one undocumented full-dump JSON endpoint (`GET /api/exposed_jobs/`) — no auth, no server-side search (query params confirmed byte-identical no-ops live), and no per-job detail endpoint (confirmed 404). Client-side filtering + dump-fetch detail resolution mirror the shape already used by the Remotive and We Work Remotely providers.
- Job IDs are parsed from the numeric segment of each job's outbound URL (`/job/go/1734670/` → `"1734670"`) rather than array index — the [ever-jobs case study](case-study/ever-jobs/packages/plugins/source-workingnomads) this was cross-checked against uses index-based IDs, which silently reassign identity when the dump reorders between fetches.
- Fixtures were captured live and replay via `hurl --test` (happy-path dump, no-op query-param proof, 404 proving no per-job endpoint).

## Scope

MCP wiring (dedicated `workingnomads_search_jobs` / `workingnomads_get_job_detail` tools, since this is a single aggregator board rather than a multi-tenant ATS) is intentionally deferred to a follow-up, per request.

## Test plan

- [x] `go build ./...` / `go vet ./...` clean
- [x] `go test ./internal/provider/workingnomads/...` passing
- [x] `hurl --test` against all 3 fixtures succeeds live
- [x] `go run ./cmd/workingnomads search|detail` smoke-tested live, including not-found and argument-validation paths


cmd/workingnomads/main.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func summarize(j workingnomads.Job) jobSummaryJSON {
+	return jobSummaryJSON{
+		ID:       j.ID,
+		Title:    j.Title,
+		Company:  j.Company,
+		Category: j.Category,
+		Location: j.Location,
+		PostedAt: j.PostedAt.Format(time.RFC3339),
```
amikai: `time.Time.Format` never returns an empty string, so a zero `PostedAt` (unparseable/missing `pub_date`) renders as `0001-01-01T00:00:00Z` — the `omitempty` tag on line 129 and the `if s.PostedAt != ""` guard on line 192 can never fire. Guard with `j.PostedAt.IsZero()` before formatting to get the intended omit behavior.

amikai: Fixed in e5fb326 — both the summarize and printDetail paths now guard on IsZero.

PR #228, opened 2026-07-18
feat(remotefirstjobs): add provider package and debug CLI

amikai:
Adds `internal/provider/remotefirstjobs` (OpenAPI + ogen client, fixture-replaying tests) and `cmd/remotefirstjobs`. MCP wiring is deferred to a follow-up session on purpose — handoff notes in the plan doc.

The ever-jobs reference scraper reads `remote-jobs.rss`, but that URL now 301s and the feed carries only five fields; the site's official public JSON API (`GET /api/search-jobs`, documented at remotefirstjobs.com/jobs-api) has server-side full-text search, category filter, and pagination, so the provider is built on that instead.

Reviewer notes:

- No per-job endpoint upstream (probed variants 404); search hits carry the full HTML description, so `Client.FindJob` scans search pages (≤500 newest per query) with optional query/category narrowing.
- Quirks verified live and documented in `openapi.yaml`: zero hits → 200 with `jobs: null`; bad category/page → 400 `{kind, message}`; naive `published_at` kept a plain string.
- All 6 hurl fixtures replay against the live API; CLI smoke-tested live.


cmd/remotefirstjobs/main.go
```diff
@@ -0,0 +1,293 @@
...
+		return enc.Encode(j)
+	}
+
+	fmt.Println(j.Title)
+	fmt.Printf("Company: %s\n", j.CompanyName)
+	fmt.Printf("Category: %s (%s)\n", j.Category, j.Seniority)
+	if len(j.Locations) > 0 {
+		fmt.Printf("Locations: %v\n", j.Locations)
+	}
+	if min, ok := j.SalaryMin.Get(); ok {
```
amikai: This gates only on `SalaryMin`, but `salary_min`/`salary_max` are independently nullable: a max-only job prints nothing, and a min-only job prints `Salary: 180000 - 0`. The search path (lines 231-232) guards on both bounds — consider matching it here.

amikai: Fixed in 9be710b: detail now guards on both bounds like the search path.

PR #231, opened 2026-07-18
feat(apple): support all public search filters in spec and client

amikai:
The Apple spec modeled only a single-country location filter, but clicking through the jobs.apple.com filter panel (and capturing the live search payloads) showed the API also accepts keyword chips, team/sub-team pairs, products, languages, a home-office flag, multi-location arrays at any granularity (state/metro/city codes such as `TPEI`/`TAIP`/`NTC9`), and four extra sort orders from the site bundle's `SEARCH_SORT_TERMS`.

**Commit 1 — spec and client:**
- `openapi.yaml`: add the missing `SearchFilters` fields and `SearchTeamFilter`, drop the wrong `maxItems: 1`/3-letter-country constraint on `locations`, extend the `sort` enum, and note the refData ID sources.
- `client.go`: expose the new filters on `SearchRequest` with bare-code inputs (`HRDWR`/`CAM`, `IPHN`, `en_US`) validated and prefixed internally; `homeOffice` is only sent when true, matching the site.
- CLI: `--home-office`, `--filter-keyword`, `--team TEAM/SUB`, `--product`, `--language` (repeatable) and the extended `--sort` enum.
- Fixtures: `jobs_filtered_req.hurl` now exercises every filter dimension in one request and its response was re-recorded live (63 matches; verified identical through the CLI).

**Commit 2 — MCP discovery tool:**
- Team/product codes are opaque taxonomy an LLM can't guess, and teams drift with Apple's org chart, so they're exposed via a new `apple_get_search_filters` tool instead of hardcoded schema enums. Teams are fetched live from the anonymous `refData/teamsofinterest` endpoint (added to the spec as `listTeams`); products are a package snapshot because that refData endpoint rejects anonymous sessions.
- `apple_search_jobs` accepts `home_office`, `keywords`, `teams` (`TEAM/SUBTEAM`), `products`, `languages`, and the full sort enum.
- Debug CLI gains a matching `filters` subcommand; verified live (`filters` → `--team MLAI/CV` returns 84 US matches, 0 in TWN, both correct).


internal/provider/apple/client.go
```diff
@@ -28,15 +28,52 @@ const (
...
+	return TeamFilter{TeamCode: teamCode, SubTeamCode: subTeamCode}, nil
+}
+
 // SearchRequest contains the stable, caller-facing Apple search parameters.
-// CountryCode is an ISO 3166-1 alpha-3 code such as TWN or USA.
+// CountryCode is an ISO 3166-1 alpha-3 code such as TWN or USA; Keyword and
+// CountryCode are required and every other filter narrows the result set.
 type SearchRequest struct {
 	Keyword     string
 	CountryCode string
```
amikai, line 66: **[P2] Expose the multi-location filters through `JobsClient`.** The OpenAPI schema now accepts multiple location IDs at country/state/metro/city granularity, but the session-aware wrapper still exposes only one required `CountryCode`. The CLI and MCP inputs inherit that restriction, so callers cannot use the multi-location behavior advertised by this PR. Please add a location input (with `CountryCode` as a compatibility fallback if desired), surface it through the CLI/MCP tool, and cover a request containing more than one location.

internal/provider/apple/products.go
```diff
@@ -0,0 +1,45 @@
+package apple
+
+// Product is one Apple products-and-services filter option.
+type Product struct {
+	Code string
```
amikai: **[P3] Keep the JSON field names consistent.** Because these fields have no JSON tags, `apple filters --format json` emits product entries as `{"Code": ..., "Name": ...}`, while the team objects and MCP output use lowercase keys. Please add `json:"code"` and `json:"name"` tags (or serialize through a normalized output type) so the machine-readable output has a consistent contract.

amikai (commented):
I found one functional coverage gap and two smaller CLI contract issues. Details are inline.

cmd/apple/main.go
```diff
@@ -82,6 +95,25 @@ func main() {
...
+				return fmt.Errorf("filters takes no positional arguments, got %v", args)
+			}
+			return runFilters(ctx, filterFlags{
+				baseURL: *baseURL,
+				timeout: *timeout,
+				format:  *format,
+			}, os.Stdout)
+		},
+	}
+	rootCmd.Subcommands = append(rootCmd.Subcommands, filtersCmd)
```
amikai, line 117: **[P3] Include `filters` in the root usage and validation error.** The new command is registered here, but the root synopsis still says `<search|detail>` and the no-subcommand error says only `search or detail`. Please update both strings so the CLI advertises all valid subcommands.

internal/provider/apple/client.go
amikai: Fixed in e2fda16: `SearchRequest` now has a `Locations []string` field (bare codes at any granularity, e.g. `TPEI`/`NTC9`) that combines with `CountryCode` via OR; `CountryCode` is now optional as long as one of the two is set. Surfaced via `--location` (repeatable) on the CLI and `locations` on the MCP tool. Covered by a mock-server round trip with two locations and no country (`TestSearchJobsMultipleLocationsWithoutCountry`, `TestAppleSearchJobsLocationsWithoutCountryE2E`) plus live verification.

internal/provider/apple/products.go
amikai: Fixed in e2fda16: added `json:"code"`/`json:"name"` tags to `Product`, so `apple filters --format json` now emits lowercase keys consistent with the team objects.

cmd/apple/main.go
amikai: Fixed in e2fda16: root usage is now `apple [FLAGS] <search|detail|filters> [FLAGS]` and the no-subcommand error says "a subcommand (search, detail, or filters) is required".

internal/provider/apple/client.go
amikai: **[P2] Follow-up: mixed-case location codes are still broken.** `locationFilterIDs` routes locations through `filterCode`, which uppercases them. Apple’s live location typeahead returns case-sensitive values such as `state953`; a direct search with `postLocation-state953` returned 2,992 matches, while `postLocation-STATE953` returned zero. Please preserve location-code case with a dedicated validator, change the OpenAPI location pattern to allow lowercase (`[A-Za-z0-9]+`), and add a regression test using a mixed-case code such as `state953`. This thread should remain open until that path is fixed.

amikai: Fixed in ea247f0: added a dedicated `locationFilterCode` validator that preserves case (unlike `filterCode`, used by team/product codes, which correctly uppercases those since they're uppercase-only). Widened the OpenAPI location pattern to `postLocation-[A-Za-z0-9]+`, regenerated, and added `TestLocationFilterCodePreservesCase` plus a mixed-case entry in `TestSearchAPIRequestMultipleLocations`. Live-verified: `--location state953` returns 2,992 matches, `--location STATE953` returns 0 — confirms the fix and that it's no longer silently coerced.

PR #232, opened 2026-07-19
feat: add Meta Careers jobs provider

amikai:
Adds a Meta Careers (metacareers.com) provider: hand-written GraphQL client, fixture-replaying tests, `cmd/meta` debug CLI, and `meta_search_jobs` / `meta_get_job_detail` / `meta_get_search_filters` MCP tools wired into the server.

Why hand-written: the site has no official API — it's a Facebook Comet app whose only surface is persisted-query GraphQL (form-encoded `doc_id` calls). The anti-bot gate is shallow: the `lsd` token is presence-only (any value passes logged-out, no cookies), but `X-FB-LSD`, `Origin`, `Referer`, and the `Sec-Fetch-*` trio are mandatory. doc_ids drift on redeploy; `doc.go` documents how to re-derive them.

Reviewer notes:
- Search is server-side filtered but **unpaginated** — `page` is a no-op (verified live), every match returns in one response. The MCP tool says so and nudges toward filters.
- hurl fixtures assert on raw bytes (`bytes contains hex,…`): the endpoint labels JSON as `text/html; charset="utf-8"` and hurl rejects the quoted charset.
- Detail rich-text fields are double-encoded (JSON string containing `{"__html": …}`); handled by a custom `UnmarshalJSON`.
- Filter value lists (teams, technologies, roles, offices) are dynamic; `meta_get_search_filters` enumerates them live. The UI's Technology filter submits under the `divisions` key — the search tools expose it as `technologies`.
- Verified: `make hurl-test` on the new fixtures, full `go test -vet=all ./...`, and a live stdio MCP smoke test for both tools.


internal/openingsmcp/meta.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+func metaHTTPToMCPResponse(response *meta.SearchResponse) *metaSearchOutput {
+	featured := make(map[string]bool, len(response.FeaturedJobs))
+	for _, job := range response.FeaturedJobs {
+		featured[job.ID] = true
+	}
+	output := &metaSearchOutput{
+		Total: len(response.AllJobs),
+		Data:  make([]metaJobSummary, 0, len(response.AllJobs)),
```
amikai, line 183: **[P2] Bound the MCP search response.** An empty-input search returns every upstream match. The checked-in fixture contains 648 jobs and about 139 KB before MCP adds URLs and mirrors the structured output into `content`, so a common call can consume tens of thousands of model tokens. Please add client-side pagination/page size or enforce a result cap while preserving `Total`.

internal/provider/meta/client.go
```diff
@@ -0,0 +1,316 @@
...
+	var data searchData
+	if err := c.graphql(ctx, searchDocID, variables, &data); err != nil {
+		return nil, fmt.Errorf("search jobs: %w", err)
+	}
+	if data.JobSearch == nil {
+		return nil, errors.New("search jobs: response has no job_search_with_featured_jobs")
+	}
+	return &SearchResponse{
+		AllJobs:      data.JobSearch.AllJobs,
+		FeaturedJobs: data.JobSearch.FeaturedJobs,
```
amikai, line 168: **[P2] Featured jobs ignore search filters.** Meta's featured rail is global, not a subset of `AllJobs`: the filtered fixture has 17 matches but its featured ID is absent from all 17, and a live zero-match query behaved the same way. Returning it here violates the `SearchJobs`/`SearchResponse` contract and makes `meta --format json search` include an unrelated job. Please either intersect it with `AllJobs` or model/document it as unfiltered global content.

amikai (commented):
Two inline findings from the post-merge review.

PR #248, opened 2026-07-26
feat(mokahr): add MokaHR ATS provider and unified-search adapter

amikai:
## Summary

Adds MokaHR (`app.mokahr.com`) as a multi-tenant ATS adapter, so 30 curated companies — DeepSeek, DJI, SHEIN, Zhihu, Kimi, 万科, 恒瑞医药 and others, ~7,000 open postings — become reachable through the existing `search_jobs_by_company` tools.

## Motivation & Context

MokaHR is the ATS behind a large share of mainland-China tech and consumer careers sites, and it was the last major gap next to the Taiwan (104, Cake) and Japan (Mynavi) coverage. It publishes no developer documentation, so `internal/provider/mokahr/openapi.yaml` was derived from live traffic and doubles as the record of the surface.

## Architecture

Two upstream traits drive the design: responses are obfuscated, and text search upstream is title-only.

```mermaid
flowchart TD
    A[ats.Adapter.Search] --> B{query or location?}
    B -- neither --> C["one page: limit=20, offset<br/>upstream total is exact"]
    B -- either --> D["readMokaHRBoard<br/>page 1 alone: reports total"]
    D --> E["remaining offsets<br/>errgroup, 8 at a time"]
    E --> F["searchDump: local match over<br/>titles + full descriptions"]
    C --> G[JobsClient]
    D --> G
    E --> G
    G --> H["Transport.RoundTrip<br/>unwrap {data, necromancer}<br/>AES-128-CBC, fixed IV"]
    H --> I["ogen client decodes<br/>the plaintext openapi.yaml describes"]
```

`Transport` is where the obfuscation stops: job endpoints answer `{"data": <base64 AES-128-CBC>, "necromancer": <the key for that very response>}` under a constant IV, the aggregations endpoint answers in the clear, and both are unwrapped before the generated client sees a body. Nothing above the transport is envelope-aware.

Everything is HTTP 200 — unknown tenants (`703002`) and unknown job ids (`703015`) arrive with `success: false` — so `JobsClient` converts those into `*APIError` and the adapter maps them onto the unified teaching errors.

## Key Changes

- **Provider package** `internal/provider/mokahr`: ogen client over a hand-written spec, deobfuscating transport, 30-company roster, fixture-replaying mock server.
- **Adapter** `internal/ats/mokahr.go`: registered in `newATSRegistry`, `careersHostPatternsByAdapter`, and `cmd/verify-companies`. Slug is `<org>/<siteId>`, because a tenant owns several careers sites with different job sets and the site a bare `/social-recruitment/<org>` redirects to is often not the largest (`high-flyer` defaults to 3 postings; its real site has 35).
- **Search shape**: upstream `keyword` matches titles only — "CLIP" and "SigLIP" each sit in one description on `high-flyer` and return zero hits — so any text query reads the board and matches locally through `searchDump`. Page 1 is fetched alone since only it reports the total; the rest fan out via `errgroup` at 8. On the largest board (Vanke, 1284 postings) that is **3.6s against 16.7s sequential**, with identical results.
- **Locations**: postings name districts ("Gongshu", "Haidian") and only the location facet knows those are Hangzhou and Beijing, so every component is rendered from the facet and the workplaces are sorted — otherwise search and detail disagree about the same job, since `getJob` returns `country` in Chinese with the localized name in `countryDescription` while `listJobs` does the opposite. Some tenants return `"locations": null` on the listing; a free-text location is resolved against the facet and pushed upstream as `locationIds` (this narrows `ey` from 340 postings to 102 for Shanghai), and a board with neither locations nor facets returns an explicit error instead of a silent zero.
- **Debug CLI** `cmd/mokahr`: `companies`, `search`, `get`, `filters`.

## Reviewer Guidance

Suggested order:

1. `internal/provider/mokahr/openapi.yaml` — the document description covers the whole upstream surface and its quirks; everything else follows from it.
2. `internal/provider/mokahr/transport.go` — the obfuscation boundary.
3. `internal/ats/mokahr.go` — `Search` / `readMokaHRBoard` / `mokaHRFacets`.
4. Tests and fixtures.

Closest attention on `readMokaHRBoard`'s page arithmetic and on the facet-based location rendering; those are the two places a subtle mistake would be quiet rather than loud.

## Verification & Testing

- `go build ./...`, `go vet ./...`, `go test ./...`, `gofmt -l` — all clean. 14 adapter tests and 11 provider tests, replaying captured fixtures through the real deobfuscation path.
- `hurl --test --max-time 30 internal/provider/mokahr/testdata/*.hurl` — 10/10 live, including a second tenant captured for the null-locations and empty-aggregations shapes.
- `openapi.yaml` passes `openapi-spec-validator`.
- `go run ./cmd/verify-companies --provider mokahr` — 30/30 OK, 0 error, 0 zero-job.
- MCP stdio smoke across `search_jobs_by_company`, `get_filters_by_company`, and `get_job_detail_by_company`, addressed both by roster name and by careers URL.

## Risks & Rollout

The obfuscation envelope and the fixed IV are internal implementation details of MokaHR's frontend, not a contract. If either changes, the transport fails loudly naming the constant rather than returning wrong data, and the hurl fixtures are what would catch it first. The roster is additive — no existing provider's behavior changes.

VIPKID is deliberately absent from the roster: its display name collides with the SmartRecruiters roster, which `NewRegistry` rejects at startup, and its MokaHR site carries one posting.


internal/provider/mokahr/openapi.yaml
```diff
@@ -0,0 +1,598 @@
...
+          examples: ["7dcd6fde-84f1-4deb-890c-f1f275df0efc"]
+        title:
+          type: string
+        status:
+          type: string
+          description: Observed only as "open" on public sites.
+        jobDescription:
+          type: string
+          description: Full description as HTML.
+        locations:
```
amikai, line 337: **[bug]** `Job.locations` is documented as JSON `null` on tenants that omit workplaces from the list endpoint (vanke / "no locations" path), and the adapter explicitly supports that case (`TestMokaHRLocationUnanswerable`, mock fixtures). The schema is only `type: array` (not nullable). ogen therefore generates a decoder that always calls `d.Arr` (`oas_json_gen.go`), and jx's `Arr` requires `[` — a `null` value fails with `"[" expected`. That breaks `ListJobs` (and thus Search/dump) for the very tenants the unanswerable-location path is written for. Sibling providers model null correctly (`type: [string, "null"]` / `nullable: true`).

**Suggestion:** Mark list (and, if observed, detail) `locations` as nullable under OAS 3.1, e.g. `type: [array, "null"]`, regenerate the client, and add a provider-level test that decodes a plaintext (or fixture) body with `"locations": null` to empty/nil without error.

internal/ats/mokahr.go
```diff
@@ -0,0 +1,605 @@
...
+) ([][]mokahr.Job, error) {
+	pageSize := cmp.Or(a.dumpPageSize, mokaHRDumpPageSize)
+	req.Limit = mokahr.NewOptInt(pageSize)
+	req.NeedStat = mokahr.NewOptBool(true)
+	req.Offset = mokahr.NewOptInt(0)
+	first, err := a.client.ListJobs(ctx, req)
+	if err != nil {
+		return nil, mokaHRError(site, err)
+	}
+	total := first.JobStats.Value.Total.Or(len(first.Jobs))
```
amikai: **[bug]** `readMokaHRBoard` sets board size as `first.JobStats.Value.Total.Or(len(first.Jobs))`. `JobStats.total` is optional in the schema. If `needStat` is honored but `total` is omitted, and the first page is full (`len(first.Jobs) == pageSize`), total becomes exactly one page, `len(pages) == 1`, and the function returns only the first page. Text search / local location matching then silently under-reads the board (missed description hits, wrong totals). The short-page early exit cannot recover because the first page is full.

**Suggestion:** If `JobStats.Total` is not set and the first page is full, do not treat `len(first.Jobs)` as the board size—page until a short page (sequential or conservative parallel), or fail closed with an error asking the caller to retry / narrow filters. Keep the current total-based parallel path only when `Total` is present.

internal/provider/mokahr/mocksrv.go
```diff
@@ -0,0 +1,139 @@
...
+// as MokaHR sent them — still obfuscated — so a client talking to this server
+// exercises the real deobfuscation path.
+//
+// Every endpoint is a POST to a fixed path, so requests are routed on their
+// body rather than on the URL.
+func NewMockServer() *httptest.Server {
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("POST /api/outer/ats-apply/website/jobs/v2", func(w http.ResponseWriter, r *http.Request) {
+		req := decodeMockRequest(r)
```
amikai, line 90: **[suggestion]** The jobs mock routes only on org/site, keyword, and offset. It ignores `locationIds` and `zhinengIds`. Adapter tests such as `TestMokaHRFilters` only assert that facet labels are accepted client-side (`facets.apply`), not that Search actually narrows upstream. A regression that drops `LocationIds`/`ZhinengIds` from the request would still pass.

**Suggestion:** Decode those fields in `mockRequest` and, when present, return a reduced/empty job set (or a dedicated fixture) so filter-driven Search assertions can check totals or membership.

internal/provider/mokahr/client_test.go
```diff
@@ -0,0 +1,215 @@
...
+func testClient(t *testing.T) *JobsClient {
+	t.Helper()
+	srv := NewMockServer()
+	t.Cleanup(srv.Close)
+	client, err := NewJobsClient(srv.URL, nil)
+	require.NoError(t, err)
+	return client
+}
+
+func TestListJobs(t *testing.T) {
```
amikai, line 22: **[suggestion]** Provider tests cover happy path, pagination, keyword, unknown site, getJob, and filter aggregations, but not the no-locations tenant fixture (`jobs_no_locations_rsp.json`) at the client layer. The only coverage is the higher-level ATS test, so a decode failure on `"locations": null` is easier to misattribute.

**Suggestion:** Add `TestListJobsNoLocations` that lists `MockNoLocationOrgID`/`MockNoLocationSiteID` and asserts non-empty jobs with nil/empty `Locations` and no error.

internal/ats/registry.go
```diff
@@ -52,6 +52,7 @@ var careersHostPatternsByAdapter = map[string]string{
 	"icims":           "careers-<slug>.icims.com/jobs/search",
 	"oracle":          "<fusion>.oraclecloud.com/hcmUI/CandidateExperience/<lang>/sites/<site>/jobs",
 	"join":            "join.com/companies/<company> (roster companies only)",
+	"mokahr":          "app.mokahr.com/{social-recruitment,campus_apply}/<org>/<siteId>",
```
amikai: **[nit]** The careers host pattern documents `{social-recruitment,campus_apply}`, but `mokaHRCareersURLRE` also accepts `campus-apply`, `campus_recruitment`, and `campus-recruitment`. Guidance shown to users is incomplete relative to what ParseCareersURL accepts.

**Suggestion:** Align the pattern string with the regex (or document the family as `campus[_-](apply|recruitment)`).

internal/ats/mokahr.go
```diff
@@ -0,0 +1,605 @@
...
+}
+
+// mokaHRError names the site in errors from a provider that answers HTTP 200
+// with a Chinese message for every failure, including unknown tenants.
+func mokaHRError(site mokaHRSite, err error) error {
+	var apiErr *mokahr.APIError
+	if errors.As(err, &apiErr) && apiErr.Code == mokahr.CodeSiteNotFound {
+		return fmt.Errorf("mokahr: careers site %s/%s not found upstream; check the org and site id in the careers URL", site.org, site.site)
+	}
+	return fmt.Errorf("mokahr: %q: %w", site.name, err)
```
amikai, line 675: **[nit]** `mokaHRError` wraps errors that often already start with `mokahr:` from `JobsClient` (`mokahr: list jobs for …`, `mokahr: upstream error …`), producing doubled prefixes like `mokahr: "DeepSeek": mokahr: list jobs for high-flyer/140576: …`.

**Suggestion:** Wrap with site context only for non-`APIError` paths, or strip/avoid the package prefix on the inner client errors when rewrapping.

amikai (commented):
## Summary

This PR adds a solid MokaHR multi-tenant adapter: OpenAPI-derived client, AES response deobfuscation, facet-aware location rendering, dual cheap-page vs full-board dump search, roster, mock fixtures, and good adapter tests. The overall design matches sibling providers (SmartRecruiters-style page/dump split, errgroup dump concurrency) and the careful comments around title-only keyword and district-vs-city locations look right. Dominant risks are schema nullability for list `locations` (documented as null on some tenants but not modeled that way), reliance on `jobStats.total` for dump page arithmetic, and mock coverage that does not exercise server-side filter IDs.

## Issue counts by severity

- bugs: 2
- suggestions: 2
- nits: 2

internal/provider/mokahr/openapi.yaml
amikai: Confirmed the mechanism, but the trigger is not what this describes — fixed in 9a68319.

An explicit `null` does fail exactly as predicted (`"[" expected: unexpected byte 110 'n'`, verified against a plaintext body through the real client). What the API actually sends, though, is **no `locations` key at all** — absent, not null — which already decoded to an empty slice. The "documented as null" wording was my error: it came from a probe using Python's `.get()`, which cannot tell an absent key from a null one.

Both halves are addressed: the document description now says absent rather than null, and `locations` is modeled `type: [array, "null"]` anyway on both `Job` and `JobDetail`. The reason to keep the nullability despite never having observed a null is that `ParseCareersURL` accepts tenants outside the roster, so the 30 sampled boards are not the whole population, and this ATS already varies its field set per tenant. Callers unwrap through `OptNilLocationArray.Or(nil)`.

Tests: `TestListJobsOmittedLocations` (captured tenant, key absent) and `TestListJobsTolerantOfNullLocations` (explicit null, inline server).

internal/ats/mokahr.go
amikai: Real, and the door is wider than described — fixed in 9a68319.

Verified against the live API: dropping `needStat` does **not** omit the total. Upstream still returns `jobStats`, with `total: 0`, alongside a full page of jobs. So `Total.Get()` reports `(0, true)`, `mokaHRPageCount(0, …)` is 1, and the read stops after one page — the same silent under-read, reached through a shape the "absent total" guard alone would not have caught.

`readMokaHRBoard` now treats both `!hasTotal` and `total <= 0` as unknown and hands off to `walkMokaHRBoard`, which pages sequentially until a short or empty page. The concurrent fan-out is kept for the case where the total is real, which is every roster tenant.

`testdata/jobs_zero_total_rsp.json` is a capture of that shape, and `TestMokaHRSearchWithoutUsableTotal` fails if the board read stops at the reported total.

internal/provider/mokahr/mocksrv.go
amikai: Agreed — fixed in 9a68319.

`mockRequest` now decodes `locationIds` and `zhinengIds`, and the jobs handler answers any filtered request with `testdata/jobs_filtered_rsp.json`, a capture of the same board narrowed by Ulanqab's location ids to two postings.

`TestMokaHRSearchFiltersUpstream` asserts the unfiltered board is 35 and both a `city` and a `category` filter come back as 2, so dropping either field from the request changes the result set instead of passing silently.

internal/provider/mokahr/client_test.go
amikai: Agreed — fixed in 9a68319.

Added `TestListJobsNoLocations` as `TestListJobsOmittedLocations`: it lists `MockNoLocationOrgID`/`MockNoLocationSiteID` and asserts non-empty jobs with ids and titles and an empty `Locations`, so a decode failure on that tenant surfaces at the client layer rather than only through the adapter.

internal/ats/registry.go
amikai: Agreed — fixed in 9a68319. The pattern now reads `app.mokahr.com/<social-recruitment|campus[_-]apply|campus[_-]recruitment>/<org>/<siteId>`, covering the whole family `mokaHRCareersURLRE` accepts.

internal/ats/mokahr.go
amikai: Agreed — fixed in 9a68319.

Resolved from the provider side rather than the adapter side: `internal/provider/mokahr` no longer prefixes its own errors, since both callers that surface them (`internal/ats` and `cmd/mokahr`) add their own context and two prefixes read as one doubled. `APIError.Error()` is now `upstream error 703002: 未找到对应的官网`, and the wrapped client errors keep their org/site context without the package name. That keeps the adapter as the single prefixer, matching the sibling adapters.

amikai:
## Two upstream findings worth keeping visible

All six review comments are addressed in 9a68319 and their threads are resolved, which collapses them. Two are recorded here instead, because in both cases what the code now guards against is **not** what the review described, and the difference is the part that would be repeated by anyone reading this later.

### `locations` is absent, not null

The review predicted that `"locations": null` would fail to decode against a bare `type: array`. That mechanism is real — verified through the client against a plaintext body:

```
"[" expected: unexpected byte 110 'n' at 121
```

But it is not what MokaHR sends. On the tenants that keep workplaces off the listing, the key is **omitted entirely** — there is no `locations` field at all — and an absent key already decoded to an empty slice. Confirmed on both the captured fixture and a live `vanke/36266` response: the string `locations` appears zero times.

The "documented as null" wording in `openapi.yaml` was my error, and it has a cause worth naming: the original probe read the field with Python's `.get()`, which returns `None` for an absent key and for a null one alike. Two different upstream shapes were collapsed into one wrong observation, and it went into the spec unchallenged.

`locations` is still modeled `type: [array, "null"]`. Not because a null has been observed, but because `ParseCareersURL` accepts tenants outside the roster — the 30 sampled boards are not the population — and this ATS already varies its field set per tenant.

### A useless total is `0`, not absent

The review's concern was an absent `jobStats.total` with a full first page, which would make the board look exactly one page long. Correct in shape, but the reachable version is different: dropping `needStat` does **not** omit the total. Upstream still returns `jobStats`, with `total: 0`:

```
POST /api/outer/ats-apply/website/jobs/v2
  {"orgId":"high-flyer","siteId":"140576","limit":5,"offset":0,"locale":"en-US"}
  -> "jobStats":{"orgId":"high-flyer","total":0}, 5 jobs
```

So `Total.Get()` returns `(0, true)`, and a guard written only for "not set" would have let it through — the same silent under-read, reached through a door the review's diagnosis did not describe. `readMokaHRBoard` now treats `!hasTotal` and `total <= 0` alike as unknown.

### Why this is here and not only in the commit

The commit message records what changed. Neither of these is reconstructable from the diff: that the API omits rather than nulls, and that a degraded total is `0` rather than missing, are facts about upstream that cost a live probe each to establish. `testdata/jobs_zero_total_rsp.json` is a capture of the second one.

PR #249, opened 2026-07-27
feat(herp): add HERP Career ATS provider

amikai:
Japanese startup hiring runs largely on HERP Hire and no existing adapter reaches it. This adds `herp` as a roster-based ATS adapter, wired end to end through the MCP server.

## Surface choice

`herp.careers` exposes two public JSON APIs. This uses the HERP Career media one, `/careers/api/v1/companies/{slug}` — the only one carrying salary (min/max/period), `jobPublishedAt`, structured `jobLocations`, `jobEmploymentTypeIds`, `jobRemoteworkType`, a `jobRoles` taxonomy, and the company profile (funding, valuation, headcount history, investors, directors). The older HERP Hire career-page API `/api/v1/{slug}` covers more tenants but has none of that; it is documented as the rejected alternative in `openapi.yaml`.

One response returns the whole board with no server-side search, so the adapter joins the full-dump family (greenhouse/lever/ashby/recruitee) and filters through `searchDump`.

## Two decisions worth a look

- **`isRemote` covers hybrid, not just full remote.** Hybrid outnumbers full remote roughly 4:1 on these boards; a strict reading would hide most remote-ish work from `location: "remote"`. `workplaceType: ["Remote"]` is the precise cut — the same key and label vocabulary ashby, lever, and bamboohr already use. The wider inconsistency in this dimension across adapters is tracked in #250.
- **Romanized prefecture aliases.** Upstream location text is Japanese-only, so `locations` is a superset of the displayed string: everything shown stays findable, plus a Latin alias per prefecture. Without it `location: "tokyo"` returns nothing with no error to explain why. Displayed locations stay Japanese.

## Known gaps

- Structured salary has no home in `ats.JobSummary`, so the parsed range is rendered into the job description instead of being searchable. A unified salary field would touch every adapter.
- Roster is seeded with 27 verified companies. Bulk expansion is a separate `roster:` PR and is bounded: the company-enumeration endpoint stops at an offset of 1000 of the 2,198 listed companies whatever `limit` is.

## Verification

- `go build ./...`, `go test ./...` green; `openapi-spec-validator` OK (the Makefile target needs Docker); `hurl --test` 3/3 live
- `verify-companies --provider herp`: 27 OK, 0 error
- Live MCP stdio: `search_jobs_by_company`, `get_filters_by_company`, `get_job_detail_by_company`, careers-URL resolution, and both error paths
- Pagination checked for overlap and stable ordering; `location: "tokyo"` and `location: "東京都"` return identical id sequences across all pages

Design and plan under `docs/superpowers/`.


internal/ats/herp.go
```diff
@@ -0,0 +1,476 @@
...
+		loc := herpLocation(&j)
+		posted, postedDate := herpParseTime(j.JobPublishedAt)
+
+		jobs = append(jobs, dumpJob{
+			summary: JobSummary{
+				JobID:    j.ID,
+				Title:    j.Name,
+				Location: loc.display,
+				PostedAt: postedDate,
+				URL:      fmt.Sprintf("%s/careers/companies/%s/jobs/%s", a.baseURL, slug, j.ID),
```
amikai: [P1] Return an application-capable job URL

`companyIsApplicationEnabled` only controls the HERP Career media surface. The current curated `unifa` response has this field set to `false`: this `/careers/companies/unifa/jobs/FfCSxynEKKev` URL ends at a page with no application action, while `https://herp.careers/v1/unifa/FfCSxynEKKev` still shows “Apply for this job”. Since `JobSummary.URL` is the applicant handoff, please select the `/v1/{slug}/{id}` URL when media applications are disabled (or use that URL consistently); otherwise valid search results lead to a dead-end page.

internal/ats/herp.go
```diff
@@ -0,0 +1,476 @@
...
+	writeHerpSection(&b, "応募受付", herpApplicationStatus(board, j))
+	return b.String()
+}
+
+func herpApplicationStatus(board *herp.CompanyBoard, j *herp.Job) string {
+	applicable, ok := j.IsApplicable.Get()
+	if !ok {
+		return ""
+	}
+	if applicable && board.CompanyIsApplicationEnabled.Or(true) {
```
amikai: [P1] Keep media opt-in separate from job applicability

These flags do not represent the same state. On that same live `unifa` job, `companyIsApplicationEnabled` is `false` but `job.isApplicable` is `true` and the `/v1` page accepts applications. This condition therefore renders an active opening as `応募受付\n停止中`. Please derive the job's open/closed status from `isApplicable`; use the company flag only to choose which application surface/URL to expose.

internal/ats/herp.go
```diff
@@ -0,0 +1,476 @@
...
+		return ""
+	}
+	return strconv.FormatFloat(amount, 'f', -1, 64) + "億円"
+}
+
+func herpDirectors(board *herp.CompanyBoard) string {
+	names := make([]string, 0, len(board.Directors))
+	for _, d := range board.Directors {
+		name := d.DirectorName
+		if slices.ContainsFunc(d.Histories, func(h herp.DirectorHistory) bool { return h.IsFounder }) {
```
amikai: [P2] Restrict the founder badge to this company

`isFounder` belongs to an individual history entry, which may describe a different company. In the current curated `kaminashi` response, 原 康紘 has `isFounder: true` for TRIDENT with `isInside: false`, so this predicate labels him as a founder of カミナシ. Please require the founder history to be inside the current company (for example, `h.IsFounder && h.IsInside`) so the detail text does not misattribute founders.

cmd/herp/main.go
```diff
@@ -0,0 +1,309 @@
...
+	if err != nil {
+		return err
+	}
+
+	matched := make([]jobSummaryJSON, 0, len(board.Jobs))
+	for _, j := range board.Jobs {
+		if f.keyword != "" && !strings.Contains(strings.ToLower(j.Name), strings.ToLower(f.keyword)) {
+			continue
+		}
+		matched = append(matched, summarize(f.company, &j))
```
amikai: [P2] Render URLs with the normalized company slug

`fetchBoard` lowercases only its local copy, so `f.company` still preserves the caller's casing here and in `runGet`. `herp --company NOTAINC search` succeeds, but every rendered URL contains `/NOTAINC/` and returns 404 (the lowercase path returns 200). Please use `board.CompanySlug` for output URLs, or normalize the slug once before both fetching and rendering.

amikai (commented):
Found three reproducible behavior issues in the new HERP integration. The implementation and full test suite otherwise look solid; details and concrete live examples are inline.

internal/ats/herp.go
```diff
@@ -0,0 +1,484 @@
...
+		out.isRemote = true
+		parts = appendDistinct(parts, label.ja)
+		aliases = appendDistinct(aliases, label.en)
+	}
+
+	out.display = strings.Join(parts, "; ")
+	// search is a superset of display: everything shown stays findable, and
+	// the romanized aliases add Latin-script reach on a board whose location
+	// text is otherwise Japanese-only.
+	out.search = strings.Join(append(parts, aliases...), "; ")
```
amikai, line 238: [P2] Keep the full upstream address in the location index

When structured locations exist, `parts` never includes `j.Location`, so `out.search` is only a superset of the shortened display value—not of the actual upstream location. The existing `ZluM2laebsdh` fixture says `京都市上京区` in `location`, but its Kyoto `jobLocations` entry has only `京都府` and a null city; an adapter search with `Location: "上京区"` therefore returns `TotalCount == 0`. Please keep the concise structured display, but append the full free-text `j.Location` to the search text so fuzzy location searches do not lose cities, stations, and addresses omitted by the structured fields.

internal/provider/herp/openapi.yaml
```diff
@@ -0,0 +1,558 @@
...
+      description: >
+        Despite the operation path, this paginates by COMPANY: `total` counts
+        companies (2,198 as of 2026-07-27) and `totalPages` is
+        `ceil(total / limit)`. Each row carries at most two preview postings
+        however many that company actually has, so this cannot stand in for a
+        cross-company job search — use it to discover slugs, then fetch each
+        board through `getCompany`.
+
+
+        Paging stops at an offset of 1000 companies whatever `limit` is
```
amikai, line 78: [P2] Regenerate after the final OpenAPI edit

At this head, `go generate ./internal/provider/herp` leaves tracked changes: 30 added lines across `oas_client_gen.go`, `oas_handlers_gen.go`, `oas_server_gen.go`, and `oas_unimplemented_gen.go`, all propagating this paging paragraph into generated documentation. Please rerun the generator and commit its output so the checked-in client is reproducible from `openapi.yaml` without an unrelated dirty diff.

amikai (commented):
Second-pass review of the updated head. The workplaceType normalization is covered and passes tests, vet, and the race detector. I found two additional reproducible issues inline; the four earlier threads also remain applicable on this head.

internal/ats/herp.go
amikai: Fixed in c3b84ff. `jobURL` now returns `/v1/{slug}/{id}` when `companyIsApplicationEnabled` is false and keeps the media URL otherwise, so the richer page stays the default for the ~all companies that accept media applications. Verified live: `/careers/companies/unifa/jobs/FfCSxynEKKev` has no apply action, `/v1/unifa/FfCSxynEKKev` has one and returns 200. `cmd/herp` applies the same rule. Pinned by `TestHerpMediaOptOutLinksToApplicableSurface` against a new real capture, `testdata/company_media_optout_rsp.json` (unifa).

internal/ats/herp.go
amikai: Fixed in c3b84ff. `herpApplicationStatus` now takes only the job and derives the status from `isApplicable` alone; the company flag is used solely by `jobURL`. `board` is no longer threaded into `herpDescription` at all, so the two facts cannot be conflated again. The unifa detail now renders `応募受付\n受付中`.

internal/ats/herp.go
amikai: Fixed in c3b84ff — the predicate is now `h.IsFounder && h.IsInside`. Confirmed against the live data you cited: kaminashi's 原 康紘 has `isFounder: true` only for TRIDENT with `isInside: false`, while 諸岡 裕人 has it for カミナシ with `isInside: true`. unifa's detail now reads `経営陣: 土岐 泰之（創業者）, 星 直人`.

cmd/herp/main.go
amikai: Fixed in c3b84ff. `summarize` and `jobURL` now take the board and render from `board.CompanySlug`, so there is a single normalized slug for both fetching and output (the report header uses it too). Verified: `herp --company NOTAINC search` emits `/careers/companies/notainc/jobs/...`, which returns 200, where the mixed-case path returns 404.

internal/ats/herp.go
amikai: Fixed in c3b84ff. The whole free-text `location`, whitespace collapsed, is now appended to the search text while the display keeps the concise structured form. `location: "上京区"` on notainc goes from 0 to 49 results through the real MCP server, and the returned `location` still reads `東京都中央区; 京都府; フルリモート`. Pinned by `TestHerpFreeTextLocationStaysSearchable`, which asserts both halves.

internal/provider/herp/openapi.yaml
amikai: Fixed in c3b84ff — you are right, the paging paragraph was added to `openapi.yaml` after the last `go generate`. Regenerated and committed the 30 lines across the four files. Re-running the generator on this head now produces a byte-identical tree.

internal/provider/herp/openapi.yaml
```diff
@@ -0,0 +1,558 @@
...
+          description: How it goes to market, one bullet per entry.
+          items:
+            type: string
+        companyPrivacyPolicyUrl:
+          type: string
+          nullable: true
+        companyIsApplicationEnabled:
+          type: boolean
+          nullable: true
+          description: Whether the company accepts applications through HERP at all.
```
amikai: [P2] Describe this as the HERP Career media flag

The implementation now correctly treats this field as choosing the application surface, but the schema still says `false` means the company does not accept applications through HERP at all. The new `unifa` capture disproves that description: this field is false while `isApplicable` is true and the `/v1` HERP Hire page accepts applications. Please update this public/generated description to say it controls applications on the HERP Career media surface; otherwise generated-client consumers are still taught the exact conflation fixed in this commit.

internal/ats/herp_test.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+func TestHerpFounderBadgeIsScopedToThisCompany(t *testing.T) {
+	a := testHerpAdapter(t)
+	d, err := a.Detail(t.Context(), herp.MockSlug, herpFullRemoteJobID)
+	require.NoError(t, err)
+
+	// isFounder rides on one history entry, which often names a previous
+	// company; only an entry with isInside refers to this one.
+	assert.Contains(t, d.Description, "経営陣: 永田 周一（創業者）")
```
amikai, line 269: [P2] Pin the outside-company founder case

This test only proves that an inside founder still gets the badge. It would also pass with the old `h.IsFounder` predicate, because this fixture has no director whose only founder history has `isInside: false`, so the regression that mislabeled kaminashi's 原 康紘 remains unguarded. Please add a negative case (a small board fixture is enough) with `{IsFounder: true, IsInside: false}` and assert that no founder badge is rendered.

amikai (commented):
Third-pass review on `c3b84ff`: all six earlier behavior issues are fixed, and the updated code passes the full suite, vet, targeted race tests, and a clean OpenAPI regeneration. Two follow-up gaps remain inline: the schema still documents the old application-flag semantics, and the founder regression test does not exercise the negative case it is intended to protect.

internal/provider/herp/openapi.yaml
amikai: Fixed in b86c0f9. The description now says the flag governs the HERP Career **media** surface, spells out that a company with it false still has a working HERP Hire career page, and points at `Job.isApplicable` as the posting-state field, citing the unifa capture. I also corrected `isApplicable` itself, which said "still accepts applications" — close enough to the same conflation to mislead — so it now states it is the posting's own state and independent of the company flag. Regenerated; `go generate` on this head is a no-op.

internal/ats/herp_test.go
amikai: Fixed in b86c0f9 — you are right, the assertion passed under either predicate. Added `testdata/company_outside_founder_rsp.json` (kaminashi): 原 康紘 has `isFounder: true` only for TRIDENT with `isInside: false`, next to 諸岡 裕人 who founded カミナシ with `isInside: true`, so one board covers both directions. The test now asserts the exact rendered line `経営陣: 原 康紘, 諸岡 裕人（創業者）, 河内 佑介` plus `NotContains "原 康紘（創業者）"`. I verified it is a real guard by reverting the predicate to `h.IsFounder`: the test fails on the NotContains assertion, and passes again with `h.IsFounder && h.IsInside`.

Worth noting for future fixtures: this shape is rare. I scanned the first 120 companies from the enumeration endpoint and none had a director whose only founder history was outside the company, which is why it needed its own capture rather than riding on an existing one.

PR #253, opened 2026-07-27
feat: add HRMOS provider and wire it into the MCP server

amikai:
Adds HRMOS 採用 (BizReach) — the highest-share ATS among mid-to-large Japanese employers — as a roster-based `ats.Adapter`, reachable through the unified `search_jobs_by_company` / `get_job_detail_by_company` / `get_filters_by_company` tools.

Its list pages embed each posting's full JD, so a whole board costs one request per 100 jobs and search needs no detail fetches. There is no server-side keyword search, so this is a full-dump adapter and `searchDump` owns filtering.

## Surface quirks worth a reviewer's attention

All are documented in `internal/provider/hrmos/doc.go`.

- **Pagination.** 100 jobs/page via `?page=N`. Past-the-end pages return **HTTP 200 with zero cassettes**, so `AllJobs` reads the max page from `nav.pg-pagenation` rather than probing for an empty page.
- **Facet job-ID lists are unusable.** They carry internal numeric IDs (`1606292663153098752`) that don't match the URL job IDs in cassette hrefs (`0000265`), and they're page-scoped. The facet *nav* itself is tenant-global and identical on every page.
- **`Filters` returns `distinctFilters(dump)`, not the facet nav.** Facet groups are tenant-defined (some tenants have none), a label can belong to two groups, and chips exist with no facet option at all. Sourcing the advertised values from the dump makes the Filters→Search round trip correct by construction, as it is for the other full-dump adapters. Consequence: `Filters` can be a strict subset of the site's own filter UI.
- **No posted date and no remote signal on the list page.** `JobSummary.PostedAt` is empty from search (`Detail` supplies it from JSON-LD), and `location=\"remote\"` matches nothing. Left as a documented gap rather than compensated for in Go code.
- **`li.sg-tag-location` can be empty.** Present exactly once per job, but 4 of 298 sampled jobs have blank text — employers who left the address off. Those jobs match no location search.

## Verification

- 10 hurl fixture pairs are real captures; `hurl --test` and `hurlfmt --check` pass against live hrmos.co, covering the multi-page stitch, the past-the-end page, both salary branches, and both 404 paths.
- Tests pin concrete values, not bands: exact job totals with no duplicate IDs, the exact facet label set, both-group and no-group chip classification, parsed salary bounds, the exact unknown-filter-key error text, and location text verbatim including the `他(N)` suffix and `［渋谷本社］` prefix.
- `go run ./cmd/verify-companies --provider hrmos` → 5 OK / 0 error against live tenants.
- Live MCP stdio smoke test passes, including combined filter + location and careers-URL resolution for non-roster tenants.

## Not included

Bulk roster curation. The seed roster is 5 verified companies; expansion is a separate pass. HRMOS has no global search and zero `/pages/` URLs in its sitemap, so slug discovery is manual.


internal/provider/hrmos/parse.go
```diff
@@ -0,0 +1,341 @@
...
+		}
+		if posting.BaseSalary != nil {
+			detail.SalaryCurrency = posting.BaseSalary.Currency
+			detail.SalaryMin = string(posting.BaseSalary.Value.MinValue)
+			detail.SalaryMax = string(posting.BaseSalary.Value.MaxValue)
+			detail.SalaryUnit = posting.BaseSalary.Value.UnitText
+		}
+	} else {
+		detail.Title = strings.TrimSpace(doc.Find("h1.sg-corporate-name").First().Text())
+		detail.Company = strings.TrimSpace(doc.Find("header.sg-breadcrumbs li").Last().Text())
```
amikai: **[P2] Read the company breadcrumb, not the job title**

`header.sg-breadcrumbs li.Last()` selects the posting title, not the company. The current test passes accidentally because this title contains `ラクスル`. `Client.JobDetail` and `cmd/hrmos` therefore report the title as `Company`; the adapter only masks this for curated roster tenants. Please read the first company breadcrumb or company-information table and assert the exact company name.

internal/provider/hrmos/parse.go
```diff
@@ -0,0 +1,341 @@
...
+		body, err := doc.Find(".pg-markdown").First().Html()
+		if err == nil {
+			detail.Description = htmlToText(body)
+		}
```
amikai: **[P2] Preserve fallback location and posting date**

The no-JSON-LD fallback leaves `Locations` and `DatePosted` empty even though the committed fixture contains the `勤務地` row and `#jsi-published-date-start`. `HrmosAdapter.Detail` consequently loses both fields for every new-graduate posting. Please parse these fallback fields from the available markup.

amikai (commented):
Two correctness issues in the new-graduate fallback path.

internal/provider/hrmos/parse.go
amikai: Fixed in 105463c.

internal/provider/hrmos/parse.go
amikai: Fixed in 105463c.

PR #279, opened 2026-07-28
feat: elicit ambiguous company selection

amikai:
## What changed

- replace `AmbiguousCompanyError` with a typed `CompanyResolution`, so multiple matches are a successful resolution state rather than an exception
- route `search_jobs_by_company`, `get_filters_by_company`, and `get_job_detail_by_company` through one shared form-elicitation flow
- present numbered, human-readable company names and public careers URLs, then continue the same logical tool call from `InputResponses`
- select the original internal adapter/slug directly after the response; no public `company_ref` or provider-qualified value such as `workday:company`
- generate the textual retry fallback only at the MCP boundary for clients without form elicitation
- migrate the production collision report and all registry callers to the new resolution contract

## Why

A duplicate company slug or name is expected data in the union of provider rosters. Modeling it as an error forced tool handlers and host models to parse teaching text, while exposing provider-qualified identifiers would leak routing details into the user contract.

The new contract separates true resolution failures from a normal choice that requires user input.

## Impact

Elicitation-capable clients can finish any ambiguous unified-company operation in one logical tool call. Users choose from company information they already understand and never need to know the underlying provider.

This intentionally breaks the internal `Registry.Resolve` API: callers now receive `CompanyResolution` and explicitly select a candidate.

Clients without elicitation still receive company names and public careers URLs for a retry.

## Validation

- `go test -count=1 ./...`
- `go vet ./...`
- `gopls check` on all changed Go files
- `git diff --check`


internal/openingsmcp/company.go
```diff
@@ -111,27 +115,193 @@ func companySearch(ctx context.Context, reg *ats.Registry, in *companySearchInpu
...
+		return 0, true, fmt.Errorf("company selection response has unexpected type %T", raw)
+	}
+
+	switch response.Action {
+	case "accept":
+		// Continue below.
+	case "decline":
+		return 0, true, errors.New("company selection declined")
+	case "cancel":
+		return 0, true, errors.New("company selection cancelled")
```
amikai: `decline` and `cancel` become bare tool errors with no next move. The host model's natural recovery from a tool error is to retry the same tool with the same arguments, which re-triggers elicitation and re-prompts a user who just said no.

Every other error string in this package teaches the recovery; these two don't. Something like "the user declined to choose a company; ask them which one they meant instead of retrying" would close the loop.

cmd/openings-mcp/main.go
```diff
@@ -52,7 +52,7 @@ Tool selection:
...
 - When the user has no target in mind, offer them the provider choices; if they don't pick one, start with the job boards (104, Cake.me, LinkedIn, Indeed, Jobindex for Denmark, and Mynavi for Japan) rather than a single company's careers site.
 - search_jobs_by_company also accepts recognized public careers-page URLs on supported ATS providers. Do not pass other careers sites; some ATS providers accept URLs only for companies already in the curated roster.
-- When a company is ambiguous, the unified company tools reject the call and list the matching companies' careers URLs; retry the same tool with one of the listed careers URLs, not with the original name.
+- When a company is ambiguous, every unified company tool asks the user to choose from human-readable company names and public careers URLs when the client supports form elicitation. Clients without elicitation receive the same choices in an error and should retry with the intended public careers URL rather than an ATS-prefixed identifier.
```
amikai: > …should retry with the intended public careers URL rather than an ATS-prefixed identifier.

The point of this PR is that `workday:nature` was never exposed. Warning the model against it plants the idea. The fallback error already states the retry rule concretely, so the trailing clause can just go.

cmd/openings-mcp/main_test.go
```diff
@@ -141,7 +154,8 @@ func TestServerListsJobTools(t *testing.T) {
 	assert.Equal(t, float64(1), companyProperty["minLength"])
 	assert.Contains(t, companyProperty["description"], "recognized public careers-page URL")
 	assert.Contains(t, companyProperty["description"], "Other careers URLs are unsupported")
-	assert.Contains(t, companyProperty["description"], "some ATS providers accept URLs only for companies in the curated roster")
+	assert.Contains(t, companyProperty["description"], "some career systems accept URLs only for companies in the curated roster")
+	assert.NotContains(t, companyProperty["description"], "ATS")
```
amikai: This bans `ATS` from tool descriptions while `serverInstructions` two lines apart still says "supported ATS providers" and "ATS-prefixed identifier" — the scrub is half-applied. Either scrub both surfaces or neither.

A bare-substring negative assertion with no stated rule behind it will also fire on any future sentence containing those three letters.

internal/ats/registry.go
```diff
@@ -70,37 +69,52 @@ var careersHostPatternsByAdapter = map[string]string{
...
+		candidates = append(candidates, candidate)
 	}
-	tw.Flush()
-	return strings.TrimRight(b.String(), "\n")
+	return candidates
+}
+
+// Select returns the adapter and provider-local slug for a candidate index.
+// Callers should select index zero directly when IsAmbiguous is false.
+func (r CompanyResolution) Select(index int) (Adapter, string, bool) {
```
amikai, line 121: Three test files each grew a near-identical helper to say "not ambiguous, give me the pair" — `selectRegistryCompany`, `selectResolvedCompany`, `selectTestCompany`. That duplication is the API asking for a companion:

```go
// Single returns the adapter and slug when exactly one company matched.
func (r CompanyResolution) Single() (Adapter, string, bool) {
	if len(r.entries) != 1 {
		return nil, "", false
	}
	return r.entries[0].adapter, r.entries[0].slug, true
}
```

That collapses all three helpers and makes the 99% caller path one call instead of an `IsAmbiguous` check plus `Select(0)`.

amikai (commented):
Reviewed against go-sdk v1.7.0. The SDK integration checks out: `oneOf` + `const`/`title` is the SDK's sanctioned titled-enum form for string properties (`mcp/client.go:996`), the capability gate matches "neither Form nor URL set means form" (`mcp/protocol.go:1830`), and the SDK skips output marshaling when `InputRequests` is set, so the pending result doesn't trip `search_jobs_by_company`'s output schema. Candidate order is deterministic across the retry (registration order, immutable registry), so the numeric choice stays valid. `go test -count=1` and `go vet` pass.

Inline comments cover the specific spots. Two things that don't anchor to a diff line:

**Spec still documents the replaced design.** `docs/superpowers/specs/2026-07-28-cross-adapter-slug-collision-options.md` still presents `AmbiguousCompanyError` as the landed decision, including a sample of the now-deleted rendered error. Worth appending a Decision note recording the move to `CompanyResolution` + elicitation.

**Test gaps.** Every error branch of `companySelection` is untested — missing response key, wrong response type, `decline`, unknown action, non-numeric or out-of-range choice. The URL-only elicitation capability (`URL != nil`, `Form == nil`) is the one non-obvious line in `supportsFormElicitation` and nothing exercises it; the fallback test uses `nil` client options, which is the no-capability case. Also, the deleted `TestCompanyToolsRejectedWhileAmbiguous` asserted "no adapter method is called while ambiguous" across all three tools; the replacement only asserts it for `get_filters_by_company`.

internal/openingsmcp/company.go
```diff
@@ -111,27 +115,193 @@ func companySearch(ctx context.Context, reg *ats.Registry, in *companySearchInpu
...
 type companyFiltersOutput struct {
 	Filters map[string][]string `json:"filters" jsonschema:"Filter dimension to its currently valid values. Pass any subset to search_jobs_by_company's filters param."`
 }
 
-func companyFilters(ctx context.Context, reg *ats.Registry, in *companyFiltersInput) (*companyFiltersOutput, error) {
-	adapter, slug, err := reg.Resolve(in.Company)
-	if err != nil {
-		return nil, err
-	}
+const companySelectionRequestID = "company_selection"
```
amikai: This const sits between `companyFiltersOutput` and `companyFilters`, unrelated to either. It belongs next to the elicitation code that uses it.

Two more small ones while you're in here: `resolveCompanyForTool` returns four values where `pending` and `err` are mutually exclusive and `adapter`/`slug` are meaningless unless both are nil — a small result struct would read better at the three call sites. And `companySelectionRequest` and `ambiguousCompanyError` both hand-render `Name — URL` with the same empty-URL branch; one `formatCandidate` helper covers both.

internal/openingsmcp/company.go
```diff
@@ -111,27 +115,203 @@ func companySearch(ctx context.Context, reg *ats.Registry, in *companySearchInpu
...
+	if selectionHint != "" {
+		fmt.Fprintf(&message, " %s", selectionHint)
+	}
+	message.WriteString(" This client cannot display a company choice form; retry with the public careers URL shown below, or the exact company name when no URL is available:")
+	for _, candidate := range candidates {
+		fmt.Fprintf(&message, "\n  %s", formatCompanyCandidate(candidate))
+		if candidate.CareersURL != "" {
+			continue
+		}
+		fmt.Fprintf(&message, " — retry with company=%q", candidate.Name)
```
amikai: The `continue` is inverted for no gain — it guards the last statement in the loop body, so the positive form says the same thing in one fewer branch:

```go
for _, candidate := range candidates {
	fmt.Fprintf(&message, "\n  %s", formatCompanyCandidate(candidate))
	if candidate.CareersURL == "" {
		fmt.Fprintf(&message, " — retry with company=%q", candidate.Name)
	}
}
```

internal/openingsmcp/company.go
```diff
@@ -111,27 +115,203 @@ func companySearch(ctx context.Context, reg *ats.Registry, in *companySearchInpu
...
+) (companyToolResolution, error) {
+	resolution, err := reg.Resolve(input)
+	if err != nil {
+		return companyToolResolution{}, err
+	}
+
+	if adapter, slug, ok := resolution.Single(); ok {
+		return companyToolResolution{adapter: adapter, slug: slug}, nil
+	}
+	if !resolution.IsAmbiguous() {
```
amikai, line 161: Worth knowing rather than changing: this branch is unreachable. `Resolve` returns an error on zero candidates, so `Single()` failing plus `!IsAmbiguous()` can never both hold — which makes this the only non-test caller of `IsAmbiguous` in production.

The refactor did improve it: the same dead check used to hide inside `Select(0)`'s `ok`, and it's now explicit. Fine to keep as a cross-package invariant guard.

cmd/openings-mcp/main.go
```diff
@@ -45,14 +45,14 @@ var (
...
 - When the user has no target in mind, offer them the provider choices; if they don't pick one, start with the job boards (104, Cake.me, LinkedIn, Indeed, Jobindex for Denmark, and Mynavi for Japan) rather than a single company's careers site.
-- search_jobs_by_company also accepts recognized public careers-page URLs on supported ATS providers. Do not pass other careers sites; some ATS providers accept URLs only for companies already in the curated roster.
-- When a company is ambiguous, the unified company tools reject the call and list the matching companies' careers URLs; retry the same tool with one of the listed careers URLs, not with the original name.
+- search_jobs_by_company also accepts recognized public careers-page URLs supported by available career systems. Do not pass other careers sites; some career systems accept URLs only for companies already in the curated roster.
```
amikai: > …recognized public careers-page URLs **supported by available career systems**.

This reads a little empty — "available career systems" gives the model no referent, where the old "on supported ATS providers" at least located the URLs somewhere. The next sentence does the real work, so the impact is small.

If you touch this again: `...URLs from the career systems this server supports` is more concrete without reintroducing the vendor category.

amikai (commented):
All seven points from the previous round are addressed — resolved those threads. Verified `go build`, `go vet`, and `go test -count=1` on the three affected packages at `4b5bb8c`, all green.

A few notes on the fixes themselves:

**The test additions go further than asked.** `TestCompanySelectionResponses` covers all eleven branches of `companySelection`. `TestURLOnlyElicitationCapabilityUsesFallback` is the standout — asserting `handlerCalled == false` proves the capability gate short-circuits *before* any elicitation is emitted, which is exactly why that `Form != nil || URL == nil` line exists. `TestAmbiguousCompanyDoesNotCallAdaptersBeforeSelection` restores the three-tool guarantee the deleted test had.

**ATS scrub is now consistent.** The remaining occurrences in `main.go` are internal identifiers (`newATSRegistry`, `atsAdapters`) and startup-error strings that never reach the model. Every user-visible surface — `serverInstructions` plus the three tool descriptions — is clean, so the rule holds without needing a `NotContains` guard to enforce it.

Three nits left inline. None block merge.

PR #281, opened 2026-07-29
feat(ats): cache full-dump boards with otter for company search

amikai:
## Summary

- Add a process-local **dump cache** so Filters → Search → Detail for the same company reuses one full-board fetch on full-dump ATS platforms.
- Backend: **[otter/v2](https://github.com/maypok86/otter)** with **TTL** (default **10m**) and **max 100** boards; simple get → load → set (no singleflight / global state).
- Inject `*DumpCache` into full-dump adapters; **`nil` means off**. CLI: `--dump-cache-ttl` (`<=0` does not construct a cache).
- **dump() returns are read-only**; BambooHR clones before filling missing list JD text for query matching.

## Scope

| In | Out |
|---|---|
| Greenhouse, Lever, Ashby, Teamtailor, Recruitee, BambooHR, Herp, Engage, Join, HRMOS, Rippling | Workday and other server-side / hybrid adapters |

## Behavior

| Concern | Behavior |
|---|---|
| Key | `{provider}:{slug}` |
| Value | `[]dumpJob` (+ optional side, e.g. Herp board) |
| Defaults | TTL 10m, max 100 entries |
| Disable | `--dump-cache-ttl=0` → inject `nil` |
| Path | `GetIfPresent` → `load(ctx)` → `Set` → return (shared, read-only) |
| Mutate | Callers that need to change fields use `cloneDumpJobs` first |

Design: `docs/superpowers/specs/2026-07-29-ats-dump-cache-design.md`.

## Test plan

- [x] `go test ./internal/ats/ -count=1`
- [x] `go test ./cmd/openings-mcp/ -count=1`
- [ ] Manual: same company Filters+Search should not re-hit upstream board twice within TTL
- [ ] Manual: `--dump-cache-ttl=0` always re-fetches


internal/ats/dump_cache.go
```diff
@@ -0,0 +1,261 @@
...
+			if err != nil {
+				return nil, err
+			}
+			// Own the slice stored in the cache; callers always get a clone.
+			return &dumpCacheEntry{
+				jobs: cloneDumpJobs(jobs),
+				side: side,
+			}, nil
+		},
+	))
```
amikai: [P2] Preserve per-request cancellation

`context.WithoutCancel` is passed directly to Otter's synchronous `Get`. The loading caller and every coalesced waiter therefore ignore their own cancellation until the upstream request finishes—30 seconds in the main binary and potentially indefinitely for other clients. Keep the shared load alive if desired, but let each waiter select on its original `ctx.Done()` and return promptly.

internal/ats/dump_cache.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func estimateDumpBytes(jobs []dumpJob) int {
+	n := 0
+	for i := range jobs {
+		n += len(jobs[i].description)
+		n += len(jobs[i].summary.Title)
+	}
+	return n
+}
```
amikai: [P2] Weight all retained dump data

The weight counts only descriptions and titles, omitting the `dumpJob` slice itself, remaining summary strings, locations, org units, cloned maps/slices, keys, and Herp's side payload. BambooHR and Engage deliberately cache description-less dumps, so they are particularly underweighted. The advertised 1 GiB bound can therefore retain several GiB of actual memory. Include all major allocations or restore hard entry/job/per-entry limits.

docs/superpowers/specs/2026-07-29-ats-dump-cache-design.md
```diff
@@ -0,0 +1,1279 @@
...
+| K6 | Do not change public `Adapter` interface | Registry, MCP tools, and tests stay stable. |
+| K7 | Integration = thin wrap of each adapter’s private `dump()` via `getOrLoadDump` | One pattern for 11 adapters; loaders stay private. |
+| K8 | **Clone on every return path**, including after singleflight waiters; `store` owns a private copy with cloned `fields` maps | Prevents BambooHR `enrichDescriptions` (and concurrent waiters) from racing or poisoning the cache. |
+| K9 | Ashby / Teamtailor / Recruitee / Herp Detail must go through the cached dump path | Otherwise Detail remains a forced miss after Search. |
+| K10 | Herp stores board metadata as optional side payload beside `[]dumpJob` | Detail needs `*herp.CompanyBoard` for company section + URL policy. |
+| K11 | Memory: max entries **32** + max jobs/entry **10_000** + coarse **MaxBytesPerEntry** on dumpJob mass only | Greenhouse `content=true` can be multi-MB; fail-open on size; tighter entry count for long sessions. Herp `side` not counted (v1). |
```
amikai: [P3] Update the committed design

This document and the PR description still promise a 2-minute TTL, `--no-dump-cache`, 32 entries, 10k jobs, and an 8 MiB cap. The final implementation instead uses 10 minutes, disables through `--dump-cache-ttl<=0`, and has a 1 GiB weighted cache. Update the design and PR body so the documented rollback command and operational limits match the shipped behavior.

amikai (commented):
Inline findings from the dump-cache review.

internal/ats/dump_cache.go
amikai: Still open on the current head: `unsafe.Sizeof(herp.Job{})` is 528 bytes, but `estimateHerpBoardBytes` charges only 128 bytes per raw job before counting any string or nested-slice backing storage. The board also retains the original Summary/skills/working-condition strings alongside the separately built `dumpJob.description`, and Directors/history payloads are omitted. The 1 GiB bound therefore remains materially underreported for Herp.

internal/ats/dump_cache.go
```diff
@@ -0,0 +1,378 @@
...
+				}
+				// Own the slice stored in the cache; callers always get a clone.
+				return &dumpCacheEntry{
+					jobs: cloneDumpJobs(jobs),
+					side: side,
+				}, nil
+			},
+		))
+		done <- dumpLoadResult{entry: entry, err: err}
+	}()
```
amikai: [P2] Avoid orphaning one goroutine per cancelled waiter

Every miss now starts a goroutine that calls `Cache.Get` with a non-cancellable context. When this caller's `ctx` is cancelled, `getOrLoadDump` returns but that goroutine remains blocked until the shared load finishes; with a client lacking `http.Client.Timeout`, it can remain forever, and a burst of cancelled waiters leaves one blocked goroutine per waiter. Keep a single bounded background load per key and let waiters select on their own contexts without each spawning an unowned goroutine; add a goroutine-leak test for a loader that never completes.

docs/superpowers/specs/2026-07-29-ats-dump-cache-design.md
```diff
@@ -0,0 +1,1301 @@
...
+| Memory bound | **MaximumWeight ≈ 1 GiB** of estimated dump payload (`DefaultDumpCacheMaxWeightBytes`) |
+| Weight model | Per-entry weigher counts dumpJob strings (including description-less boards), fields maps, struct overhead, and Herp `*CompanyBoard` side payload |
+| Coalescing | otter `Cache.Get` stampede protection |
+| Cancellation | Shared load uses `context.WithoutCancel`; each waiter still returns on its own `ctx.Done()` |
+| Clone | Per-waiter `cloneDumpJobs` (BambooHR enrich safe) |
+| CLI | Only `--dump-cache-ttl` (default `10m`) |
+
+Historical sections below may still mention 2m TTL, `--dump-cache-ttl<=0`, 32
+entries, 10k jobs, or 8 MiB per-entry caps — treat those as superseded by this
+table unless noted.
```
amikai: [P3] Correct the disable-flag replacements

This row currently says `--dump-cache-ttl<=0` is the supported syntax and then says that same syntax does not exist. The next paragraph also treats the current syntax as a superseded historical value. Similar replacements later list `--dump-cache-ttl<=0` as if it were a second flag and still reference `noDumpCache`. Say that there is no `--no-dump-cache`, then clean or remove the superseded plan text and update the still-stale PR body.

amikai (commented):
Second-pass findings after validating the review-fix commit.

docs/superpowers/specs/2026-07-29-ats-dump-cache-design.md
amikai: Still open: the PR body continues to advertise the removed `--no-dump-cache`, 2-minute TTL, and old capacity limits. The new as-shipped section also contains contradictory flag replacements; I added a current-line inline thread for those documentation errors.

docs/superpowers/specs/2026-07-29-ats-dump-cache-design.md
amikai: Still open on `4d95210`. The PR body and the authoritative “As shipped” table are now correct, but the committed design still presents removed behavior as current implementation: the overview says “otter-backed coalescing” (lines 45–48), Goals still require singleflight and hard entry/job/byte caps (lines 120–132), and the core code sample still imports `singleflight`, defines `MaxEntries`/`MaxJobsPerEntry`, and uses a 2-minute TTL (lines 247 onward). Please delete the superseded implementation plan or rewrite it to the shipped get/load/set + 10m/1GiB model; a top-level disclaimer leaves the document internally contradictory.

internal/ats/teamtailor.go
```diff
@@ -123,23 +126,25 @@ func (a *TeamtailorAdapter) Detail(
...
-	feed, err := a.feed(ctx, slug)
+	// Route through dump so Search → Detail reuses the process-local dump
+	// cache. Company is roster name or host slug (not feed.Title) for v1.
+	jobs, err := a.dump(ctx, slug)
 	if err != nil {
 		return nil, err
 	}
-	for _, j := range feed.Items {
-		if j.ID.String() != jobID {
+	company := cmp.Or(teamtailor.CompaniesByHost[strings.ToLower(slug)].Name, slug)
```
amikai, line 135: [P2] Preserve the feed's company name for uncurated hosts

`ParseCareersURL` deliberately accepts non-roster `*.teamtailor.com` sites (and curated custom domains), but this fallback now returns the raw hostname for those boards. Before this change, `Detail` used `feed.Title`—the fixture identifies `knaufsemea.teamtailor.com` as “Knauf Belgium”—whereas the updated test now blesses the hostname as the company. Carry the title in the cached side payload (or the dump) so caching does not regress the user-visible company name.

amikai (commented):
One final correctness finding from the merged-head review.

PR #292, opened 2026-08-11
feat(dayforce): add Dayforce ATS provider and adapter

amikai:
Adds Dayforce (formerly Ceridian) Recruiting as a multi-company ATS provider, reachable through the unified `search_jobs_by_company` / `get_filters_by_company` / `get_job_detail_by_company` tools.

Every Dayforce customer's candidate portal is served from one public JSON API on `jobs.dayforcehcm.com`, keyed by `clientNamespace` + `careerSiteXRefCode`, with genuine server-side search — so this is the roster-based ATS shape, not a bespoke surface. Full recon and the slice-by-slice plan are in `docs/superpowers/plans/2026-08-11-dayforce-provider.md`.

## Constraints worth knowing when reviewing

- **POST is gated by next-auth CSRF.** A bare `403 Forbidden` (9 bytes, non-JSON) regardless of body until both the `/api/auth/csrf` cookie and the `X-CSRF-TOKEN` header are present. Modeled the apple way: csrf as a spec operation, cookie jar on the `*http.Client` passed to ogen's `WithClient`, token as a typed header param. `mocksrv.go` requires both legs, so a regression fails a Go test rather than only showing up live — and it is `httptest.NewTLSServer`, because the gate cookie is `__Host-`/Secure and a plain-HTTP mock would silently drop it.
- **Page size is a hard 25**; `paginationStart` is the only pagination knob (echoed back as `offset`). `pageSize`/`count`/`page`/`skip` are silently ignored, so the adapter owns the arithmetic.
- **`jobBoardId` is not `jobBoardCode`.** Detail and the facet endpoints need the numeric id, which never appears in the human URL. It is 1 for most `CANDIDATEPORTAL` boards but `mydayforce`'s `alljobs` board is 8, so it is stored per roster row and resolvable from the SSR page's `__NEXT_DATA__` `site-info`.
- **Unsupported `cultureCode` returns 200 with `maxCount: 0`**, not an error — a wrong culture is indistinguishable from an empty board, so it is pinned per roster row.

## The one design decision that needs a reviewer's eye

`openapi.yaml` deliberately models **only the fields we consume**. ogen's decoder is strict: one field whose live value contradicts its declared type fails the entire response — a whole 25-item page, or a whole detail call. Four rounds of "widen the field that threw" each uncovered another (`isoCountryCode`, then `isoCurrencyRegion` + `postingLocations` as a null array + a `bool` missing from a `oneOf`, then `jobDescriptionFooter` + `assessmentSentType` declared `string` but actually an integer). Null-detection was the wrong invariant: a field null in every sampled tenant still has an unguessable non-null type.

So the candidate portal's own apply/assessment/render state (`assessmentId`, `assessmentSentType`, `hasApplied`, `canApplyAgain`, `jobApplicationTemplateId`, `postingThemeOverrideBoardCode`, `postingStatus`, `postingType`, `postingAppliedStatus`) is left unmodeled — ogen ignores undeclared response fields, so they cannot break decoding. This trades away the usual preference for surfacing everything upstream returns; anything dropped is one spec addition away from returning, with its type confirmed from live data rather than guessed. The reasoning and the rejected alternatives are recorded in both the plan and `info.description`.

## Verification

- Walked **every search page of 14 tenants** (roster + `corpay`, `odc`, `ipg`, `kiinc`, `medplast`, `opta`, `paradigm`, `thechronicle`) plus ≥10 detail objects each through the real client: **zero decode errors**. This is the check that matters — each breaking row was the only one of its kind on its board, so unit fixtures cannot find them.
- Live MCP stdio: all six roster companies via `search_jobs_by_company` and `get_job_detail_by_company`, `get_filters_by_company` returning its three dimensions, and a non-roster board URL (`.../en-US/corpay/CANDIDATEPORTAL`).
- `go build ./...`, `go vet ./...`, `make validate-openapi`, `make hurl-lint` clean; `go test -race ./internal/provider/dayforce/...` and `go test ./internal/ats/... ./cmd/openings-mcp/...` green; all 14 hurl fixtures pass live.

## Notes

- Seed roster is 6 companies. The API never returns an employer name, so each display name was sourced from that tenant's own live posting text; `corpay` and `kiinc` were left out of the seed because their names collide with existing workday/workable entries. Bulk roster expansion via `unverified/dayforce.yaml` is deliberately deferred to a separate `roster:` PR.
- The collision golden file gains 4 lines (dayforce↔join "AV Group", dayforce↔smartrecruiters "PCA") — expected cross-adapter overlap, resolved by careers URL.
- `testdata/siteinfo_rsp.html` is a faithful capture of a public SSR page and therefore contains Dayforce's own publicly-served Google Maps browser key. Kept verbatim per the real-captures convention.


internal/provider/dayforce/client.go
```diff
@@ -0,0 +1,212 @@
...
+	case *SearchResponse:
+		return response, nil
+	case *SearchJobPostingsForbidden:
+		// The cached token may have gone stale; refetch once before
+		// surfacing the error, so a long-lived BoardClient recovers from an
+		// expired cookie without the caller having to retry.
+		if _, refreshErr := c.csrfToken(ctx, true); refreshErr != nil {
+			return nil, fmt.Errorf("%w: refresh token failed: %v", ErrCSRFRequired, refreshErr)
+		}
+		return nil, ErrCSRFRequired
```
amikai: **[bug]** On `403`, `Search` refreshes the cached CSRF token and then returns `ErrCSRFRequired` without retrying the POST. The comment on lines 130–132 says a long-lived `BoardClient` recovers from an expired cookie “without the caller having to retry,” but `DayforceAdapter.Search` does not retry either. `TestSearchRefreshesTokenOnForbidden` confirms the first call after a stale token fails and only the next call succeeds. A long-lived MCP process will therefore fail the first search after cookie/token expiry (or a first `GetCsrf` that stored a token without a usable cookie) even though the client already holds a fresh pair.

**Suggestion:** After a successful `csrfToken(ctx, true)`, retry `SearchJobPostings` once with the new token and return that result. Keep a single retry so a persistent Cloudflare/non-CSRF `403` still surfaces. Update the test to expect the stale-token call to succeed.

internal/ats/dayforce.go
```diff
@@ -0,0 +1,493 @@
...
+	if c, ok := dayforce.CompaniesBySlug[strings.ToLower(slug)]; ok {
+		return c.Name, c, nil
+	}
+
+	ns, xref, ok := strings.Cut(slug, "/")
+	if !ok || ns == "" || xref == "" {
+		return "", dayforce.Company{}, fmt.Errorf("dayforce: unknown company %q; pass a roster slug or a jobs.dayforcehcm.com board URL", slug)
+	}
+
+	info, err := a.client.SiteInfo(ctx, ns, xref, dayforceDefaultCulture)
```
amikai: **[bug]** `parseDayforceCareersURL` treats the culture segment as a non-capturing group and `resolveSlug` always calls `SiteInfo` / search / detail with `dayforceDefaultCulture` (`en-US`). The spec and adapter comments document that an unsupported culture returns `200` with `maxCount: 0` (indistinguishable from an empty board), and that SiteInfo `404`s on a culture the board does not serve. For a URL like `https://jobs.dayforcehcm.com/fr-CA/{ns}/{xref}`, the only reliable culture signal is thrown away: SiteInfo is fetched as `/en-US/{ns}/{xref}` (404 if that locale has no page), and search is pinned to `en-US` (silent empty if that locale has no jobs). Roster rows can override culture; URL-resolved boards cannot.

**Suggestion:** Capture the culture from `dayforceBoardURLRE` / `dayforceLegacyURLRE` when present, persist it on the minted `{ns}/{xref}` slug or on the resolved `Company`, and pass it to `SiteInfo` and subsequent API calls. Keep `en-US` only for locale-less URLs and roster rows that omit `culture_code`.

internal/ats/dayforce.go
```diff
@@ -0,0 +1,493 @@
...
+		JobBoardCode:    board.JobBoardCode,
+		CultureCode:     board.Culture(),
+		PaginationStart: dayforce.NewOptInt(pageIndex * dayforcePageSize),
+	}
+	if p.Query != "" {
+		req.SearchText = dayforce.NewOptString(p.Query)
+	}
+	if location := strings.TrimSpace(p.Location); location != "" {
+		req.LocationString = dayforce.NewOptString(location)
+		req.Distance = dayforce.NewOptFloat64(dayforceDefaultDistance)
```
amikai, line 332: **[suggestion]** When `SearchParams.Location` is set, the adapter sends `locationString` and `distance: 50` but omits `distanceUnit` (and the `locationId`/`locationType` zeros the live portal sends). The recon payload that verified server-side location narrowing used `distance: 50, distanceUnit: 0` together. If the API ignores `distance` without a unit, location search silently degrades to a weaker or no-op filter. The debug CLI (`cmd/dayforce/main.go`) does not set `distance` at all, so it cannot be used to confirm the same path.

**Suggestion:** Set `DistanceUnit` to the observed `0` whenever `Distance` is set (and consider sending the same `locationId`/`locationType` zeros the portal uses, if still required). Mirror that in the CLI so live debugging matches the adapter.

internal/provider/dayforce/openapi.yaml
```diff
@@ -0,0 +1,615 @@
...
+        maxCount:
+          type: integer
+        offset:
+          type: integer
+        count:
+          type: integer
+
+    SearchPosting:
+      type: object
+      required:
```
amikai, line 361: **[suggestion]** The spec’s own decode-strictness note says to declare only consumed fields, because one mistyped required field fails a whole 25-row page. `SearchPosting` still requires unused fields (`jobReqId`, `jobDescription`, `searchScore`, `isEvergreen`, `reapplyPeriodDays`, `clientNamespace`, `jobBoardId`, …), and `JobPostingDetail` still requires unused ones (`isInternal`, `relocationEligible`, `createdTimestampUTC`, `isoCurrencyRegion`, `jobPostingAttributes`, …). That is the same class of defect that already broke decoding on null locations, null currency, null footer, and bool attributes.

**Suggestion:** Keep required only the fields the adapter/CLI actually read (`jobPostingId`, `jobTitle`, `hasVirtualLocation`, `postingStartTimestampUTC`, `postingLocations`, `jobPostingContent`, …). Leave everything else undeclared until a caller needs it with a live-pinned type.

internal/provider/dayforce/siteinfo.go
```diff
@@ -0,0 +1,106 @@
...
+
+// SiteInfo is a board's identity, resolved from the SSR candidate portal
+// page's embedded Next.js state rather than a JSON endpoint — there isn't
+// one. It carries JobBoardId, which detail and posting-attribute calls need
+// but which never appears in the page's own URL.
+type SiteInfo struct {
+	ClientNamespace string `json:"clientNamespace"`
+	JobBoardCode    string `json:"jobBoardCode"`
+	CultureCode     string `json:"cultureCode"`
+	ClientId        int    `json:"clientId"`
```
amikai: **[suggestion]** `SiteInfo` unmarshals unused `ClientId int` and never checks `JobBoardId > 0`. `clientId` is not read by the adapter; a tenant that returns `"clientId": null` (the same null-vs-int pattern already seen on job payloads) would fail `parseSiteInfo` and block every URL-resolved Search/Filters/Detail, even though Search does not need `jobBoardId`. A missing/zero `jobBoardId` currently succeeds and fails later on detail/attribute calls with a poorer error.

**Suggestion:** Drop `ClientId` (and any other unused site-info field). Require `JobBoardId > 0` before returning success, matching roster validation.

internal/provider/dayforce/mocksrv.go
```diff
@@ -0,0 +1,226 @@
...
+	// MockNullCurrencyJobID is detail_null_currency_rsp.json's posting id
+	// (jdemea 35762), whose detail response has isoCurrencyRegion: null and
+	// hasVirtualLocation: false with zero postingLocations — for tests
+	// exercising [BoardClient.Job] against that shape.
+	MockNullCurrencyJobID = 35762
+
+	// MockBoolAttributeJobID is detail_bool_attribute_rsp.json's posting id
+	// (mymilacron 2684), whose jobPostingAttributes carries a
+	// TravelRequired attribute with a boolean value and type "bool" —
+	// discovered live during the page-walk this fix's evidence sweep did
```
amikai: **[suggestion]** Several new comments narrate design history rather than a non-obvious constraint: the `MockBoolAttributeJobID` comment recounts the page-walk that found the bool attribute; `BoardClient`’s doc comment explains why the type is not named `Client`; `Search`’s 403 comment describes the intended recovery story (and is also inaccurate — see the CSRF retry issue). Repo convention is concise WHY, not change narrative.

**Suggestion:** Trim to the constraint a later reader needs (e.g. “TravelRequired is a JSON bool with type `bool`”). Move naming/history to the plan doc if it must be kept.

amikai (commented):
## Summary

This is a complete, well-structured Dayforce ATS integration: ogen spec + CSRF-aware `BoardClient`, roster, debug CLI, MCP adapter, and registry/verify-companies wiring all follow existing provider conventions (UltiPro-style non-roster slugs, Apple-style cloned client + cookie jar, collision golden updates). The live-derived fixtures and typed 403/404 mapping are thorough. The remaining risks are CSRF recovery that still fails the first search after a stale token, culture being discarded from board URLs (the silent-empty trap they documented), and leftover strict-decode surface on unused required fields.

## Issue counts by severity

- bugs: 2
- suggestions: 4
- nits: 0

amikai (commented):
Reviewed the hand-written code (adapter, client, siteinfo, roster, CLI, spec) — `go build`, `go vet`, `go test ./internal/ats/... ./internal/provider/dayforce/... ./cmd/openings-mcp/...` all pass on this branch, the collision golden regenerates clean, and the CSRF gate / null-shape / pagination paths are well covered. No dropped errors, nil derefs, or broken callers found. Six inline notes, roughly in severity order; the first is the only one I'd call a real disagreement with the rest of the repo.

Two things I checked and am **not** flagging: the Google Maps browser key in `siteinfo_rsp.html` matches the existing real-capture convention (`google`, `hrmos` fixtures carry the same kind of public key), and `distanceUnit: 0` is not a guess — the plan doc records it from the portal's own request body.

internal/ats/dayforce.go
```diff
@@ -0,0 +1,513 @@
...
+// with (tests point the client at a mock server but still render the real
+// public host, matching smartrecruiters.RosterCompany.CareersURL).
+const dayforceJobsHost = "https://jobs.dayforcehcm.com"
+
+// dayforcePageSize is Dayforce's fixed, non-negotiable search page size
+// (verified live — see internal/provider/dayforce openapi.yaml). It is
+// deliberately not the shared ats.pageSize: that constant is the uniform
+// value every full-dump and re-paged adapter presents, but Dayforce search
+// is genuinely server-side against a page size this adapter cannot change.
+const dayforcePageSize = 25
```
amikai: **25-row pages break the unified 20-job contract.**

`search_jobs_by_company` documents its `page` param as *"1-based page number; each page returns at most 20 jobs"* (`internal/openingsmcp/company.go:49`), and every other server-side adapter requests exactly `ats.pageSize`. Here page 1 of `pca` returns 25 rows and `total_pages = ceil(352/25) = 15`, so this one provider answers with a different window size and a different page-count denominator than the rest of the roster.

The comment says the page size "this adapter cannot change" — true upstream, but the adapter *can* still present the uniform page. `internal/ats/icims.go:163-195` is exactly this case (fixed, non-negotiable upstream size, sometimes >20) and maps it onto 20 with

```go
correctPr := start / upSize
offsetInPage := start % upSize
// … slice from offsetInPage, stitch the next upstream page while len(selected) < pageSize
```

Two ways out, either is fine — re-page onto 20 the iCIMS way (≤2 upstream requests per unified page), or keep 25 and amend the `page` description in `company.go` so the tool contract stops promising 20. Right now the schema and this adapter disagree.

internal/ats/dayforce.go
```diff
@@ -0,0 +1,513 @@
...
+	host := strings.ToLower(u.Hostname())
+	subject := host + u.EscapedPath()
+
+	if m := dayforceBoardURLRE.FindStringSubmatch(subject); m != nil {
+		ns = namedGroup(dayforceBoardURLRE, m, "ns")
+		xref = namedGroup(dayforceBoardURLRE, m, "xref")
+		culture = namedGroup(dayforceBoardURLRE, m, "culture")
+		if dayforceReservedFirstSegments[strings.ToLower(ns)] {
+			return "", "", "", false
+		}
```
amikai, line 188: **The optional culture group gets swallowed as the namespace, and this guard can't see it.**

`(?:(?P<culture>…)/)?` is optional and the rest of the pattern needs only two segments, so a `/{culture}/{ns}` path backtracks into `ns=<culture>`. Verified by running the adapter:

| URL | slug |
| --- | --- |
| `/en-US/pca` | `en-us/pca` |
| `/en-US/pca/jobs/62374` | `en-us/pca` |
| `/en-US/profile` | `en-us/profile` |

The last row also defeats `dayforceReservedFirstSegments`, whose doc says those prefixes "must never be mistaken for a tenant namespace" — it only inspects the `ns` group, which the culture segment has displaced.

`Registry.Resolve` returns on the first adapter that claims a URL, so instead of the teaching error (*"unrecognized careers URL …; supported careers-page hosts: …"*) the caller gets a wasted SSR fetch of `/en-US/en-US/pca` and `dayforce: resolve board "en-us/pca": fetch dayforce site-info page: unexpected status 404`.

```suggestion
		if dayforceReservedFirstSegments[strings.ToLower(ns)] {
			return "", "", "", false
		}
		// A two-segment path is ambiguous: the optional culture group
		// backtracks into ns, so /en-US/pca parses as ns "en-US". A
		// culture-shaped namespace is far less likely than a truncated board
		// URL, so reject rather than mint a slug that can only 404.
		if culture == "" && dayforceCultureRE.MatchString(ns) {
			return "", "", "", false
		}
```

(pairs with the `dayforceCultureRE` suggestion below)

internal/ats/dayforce.go
```diff
@@ -0,0 +1,513 @@
...
+	return "", "", "", false
+}
+
+func canonicalDayforceCulture(culture string) string {
+	if !regexp.MustCompile(`(?i)^[a-z]{2}-[a-z]{2}$`).MatchString(culture) {
+		return ""
+	}
+	parts := strings.SplitN(culture, "-", 2)
+	return strings.ToLower(parts[0]) + "-" + strings.ToUpper(parts[1])
+}
```
amikai, line 218: **Regexp compiled on every call.** The file's other two patterns are package-level `MustCompile`; this one recompiles per call — and `Registry.Resolve` polls every adapter for URL-shaped input, hitting this twice per matching URL (`parseDayforceCareersURL` + `ParseCareersURL`) plus once per non-roster `resolveSlug`.

```suggestion
// dayforceCultureRE matches one culture segment, e.g. "en-US".
var dayforceCultureRE = regexp.MustCompile(`(?i)^[a-z]{2}-[a-z]{2}$`)

func canonicalDayforceCulture(culture string) string {
	if !dayforceCultureRE.MatchString(culture) {
		return ""
	}
	parts := strings.SplitN(culture, "-", 2)
	return strings.ToLower(parts[0]) + "-" + strings.ToUpper(parts[1])
}
```

internal/ats/dayforce.go
```diff
@@ -0,0 +1,513 @@
...
+	ns, xref := parts[0], parts[1]
+	culture := dayforceDefaultCulture
+	if len(parts) == 3 {
+		culture = canonicalDayforceCulture(parts[2])
+		if culture == "" {
+			return "", dayforce.Company{}, fmt.Errorf("dayforce: unknown company %q; pass a roster slug or a jobs.dayforcehcm.com board URL", slug)
+		}
+	}
+
+	info, err := a.client.SiteInfo(ctx, ns, xref, culture)
```
amikai: **Non-roster boards pay a ~431 KB HTML fetch per call for a field search doesn't use.**

`resolveSlug` always resolves `jobBoardId` through the SSR page, but `Search` only needs it when `p.Filters` is non-empty (`applySearchFilters` → `resolveAttribute`); the search body itself carries no `jobBoardId`, and every search row returns one. Nothing caches the result either, so a filter-less `search_jobs_by_company` against `.../en-US/corpay/CANDIDATEPORTAL` downloads and goquery-parses `testdata/siteinfo_rsp.html`-sized HTML on **every page request**, then throws the value away. It also makes the JSON-only search path fail outright on any `__NEXT_DATA__` shape drift.

Sketch: keep the culture check (the page 404s on an unsupported culture, which is the useful half of this call) but make the id lazy and memoized:

```go
// resolveSlug returns the board with JobBoardID unset for non-roster slugs;
// jobBoardID(ctx, board) fills it in from a memoized SiteInfo on the paths
// that need it (Filters, Detail, filtered Search).
```

A `sync.Map` keyed by `ns/xref/culture` would also collapse the Search→Detail pair down to one fetch.

internal/ats/dayforce.go
```diff
@@ -0,0 +1,513 @@
...
+		CultureCode:     board.Culture(),
+		PaginationStart: dayforce.NewOptInt(pageIndex * dayforcePageSize),
+	}
+	if p.Query != "" {
+		req.SearchText = dayforce.NewOptString(p.Query)
+	}
+	if location := strings.TrimSpace(p.Location); location != "" {
+		req.LocationString = dayforce.NewOptString(location)
+		req.Distance = dayforce.NewOptFloat64(dayforceDefaultDistance)
+		req.DistanceUnit = dayforce.NewOptInt(0)
```
amikai: **Only search input with no coverage.** No Go test drives `p.Location`, no hurl fixture posts a location body, and `pcaSearchFixture` branches on `searchText`/`paginationStart` only — so dropping `distanceUnit`, flipping the 50-mile default, or losing `locationString` entirely keeps `go test ./...` and the hurl suite green.

Worth adding both: a `mocksrv` branch on `req.LocationString` (fixture selection already reads the body) plus a hurl fixture replaying the payload the plan doc captured from the portal (`"distance": 50, "distanceUnit": 0`), which is also the evidence for the bare `0` here:

```suggestion
	if location := strings.TrimSpace(p.Location); location != "" {
		// distanceUnit 0 is miles — the value the portal's own search body
		// sends alongside distance (see the plan doc's captured payload).
		const distanceUnitMiles = 0
		req.LocationString = dayforce.NewOptString(location)
		req.Distance = dayforce.NewOptFloat64(dayforceDefaultDistance)
		req.DistanceUnit = dayforce.NewOptInt(distanceUnitMiles)
	}
```

internal/provider/dayforce/client_test.go
```diff
@@ -0,0 +1,502 @@
...
+	require.NoError(t, err)
+
+	assert.Equal(t, 1, info.JobBoardId)
+	assert.Equal(t, "pca", info.ClientNamespace)
+	assert.Equal(t, "candidateportal", info.JobBoardCode)
+	assert.Equal(t, "en-US", info.CultureCode)
+}
+
+// TestSiteInfoMissingNextData proves parseSiteInfo fails rather than
+// zero-valuing when the page has no __NEXT_DATA__ script at all.
```
amikai: Copy of `TestSiteInfoMissingNextData`'s comment — this fixture *is* a well-formed `__NEXT_DATA__`, with `jobBoardId: 0`. It's the only test covering the new guard in `siteinfo.go`, so the comment should say so (and it currently makes the real missing-`__NEXT_DATA__` test two functions below look redundant).

```suggestion
// TestSiteInfoMissingJobBoardID proves parseSiteInfo rejects a site-info
// query whose jobBoardId is absent or zero, rather than returning a board
// whose detail and posting-attribute calls would all address jobBoardId 0.
```

internal/ats/dayforce.go
```diff
@@ -0,0 +1,578 @@
...
+		return "", dayforce.Company{}, fmt.Errorf("dayforce: unknown company %q; pass a roster slug or a jobs.dayforcehcm.com board URL", slug)
+	}
+	ns, xref := parts[0], parts[1]
+	culture := dayforceDefaultCulture
+	if len(parts) == 3 {
+		culture = canonicalDayforceCulture(parts[2])
+		if culture == "" {
+			return "", dayforce.Company{}, fmt.Errorf("dayforce: unknown company %q; pass a roster slug or a jobs.dayforcehcm.com board URL", slug)
+		}
+	}
```
amikai, line 272: **Follow-up from the lazy SiteInfo: nothing validates the assumed culture on the search path any more.**

`resolveSlug` no longer touches the network, so for a non-roster board the `en-US` default is now taken on faith. Per the spec, an unsupported `cultureCode` returns `200` with `maxCount: 0` — the documented silent-empty trap — while the SSR page 404s. So a locale-less URL (`jobs.dayforcehcm.com/<ns>/CANDIDATEPORTAL`, the form the regex explicitly supports) for a board that does not serve `en-US` used to fail loudly on the SiteInfo 404 and now reports "no jobs found" instead.

Cheap fix that keeps the fetch off the happy path: the zero-result early return at line 324 is already the exact hook — probe there, only when the board came from a URL.

```go
if start >= first.MaxCount {
	// A non-roster board's culture is assumed, and an unsupported culture
	// returns 200/maxCount 0 rather than an error — so an empty result is
	// the one place worth paying for SiteInfo, which 404s on a culture the
	// board does not serve.
	if board.JobBoardID == 0 {
		if _, err := a.jobBoardID(ctx, &board); err != nil {
			return nil, err
		}
	}
	return &SearchResult{Jobs: []JobSummary{}, TotalCount: first.MaxCount, Page: page, TotalPages: totalPages(first.MaxCount)}, nil
}
```

Roster rows keep their pinned, verified culture and never hit this. Fine to defer — flagging so the trap the spec, `defaultCultureCode`, and `SiteInfo`'s own doc comment all warn about does not go unguarded by accident.

amikai (commented):
Re-reviewed at 91e8312. All six of my earlier points are addressed and I verified the re-paging independently: driving the adapter against a synthetic 352-job board (25/upstream page) returns pages of exactly 20, `total_pages` 18, page 18 with the remaining 12, empty past the end, and the concatenation of pages 1–18 is the upstream order with no gaps and no duplicates. `go build`, `go vet`, `go test ./...` (incl. `-race` on the provider), `make validate-openapi`, and `make hurl-lint` are all clean.

One new thing the lazy SiteInfo introduced, inline below — non-blocking.

internal/ats/dayforce.go
amikai: Fixed in 91e8312 — `Search` now remaps the upstream 25-row pages onto `ats.pageSize` the iCIMS way (`start/25`, `start%25`, stitch, truncate) and uses the shared `totalPages`. Verified against a synthetic 352-job board: pages of exactly 20, `total_pages` 18, 12 on the last page, empty past the end, and pages 1–18 concatenate to the upstream order with no gaps or duplicates.

internal/ats/dayforce.go
amikai: Fixed in 91e8312 — the culture-shaped-`ns` reject landed, and `/en-US/pca`, `/en-US/pca/jobs/62374`, and `/en-US/profile` are now in `TestDayforceParseCareersURLRejects`. The legitimate forms (locale-less `/corpay/CANDIDATEPORTAL`, `/fr-CA/{ns}/{board}`, legacy hosts) still resolve.

internal/ats/dayforce.go
amikai: Fixed in 91e8312 — hoisted to the package-level `dayforceCultureRE`, which the new two-segment guard reuses.

internal/ats/dayforce.go
amikai: Fixed in 91e8312 — non-roster rows now leave `JobBoardID` unset and `jobBoardID()` fills it from a `sync.Map`-memoized SiteInfo only on the paths that need it (`Filters`, `Detail`, filtered `Search`), so a filter-less search makes no HTML request at all and a Search→Detail pair costs one.

One consequence worth its own thread — with the fetch gone from the search path, nothing validates the assumed `en-US` for a locale-less non-roster URL: see the follow-up comment on the culture default.

internal/ats/dayforce.go
amikai: Fixed in 91e8312 — `distanceUnitMiles` names the constant in both the adapter and the CLI, `pcaSearchFixture` branches on `locationString`, `TestDayforceSearchByLocation` covers the adapter path, and `search_location_req.hurl` replays the captured payload. The hurl file lints clean; I could not run it live from here, so its live-replay status is unverified on my side.

internal/provider/dayforce/client_test.go
amikai: Fixed in 91e8312.

internal/provider/dayforce/client.go
amikai: Fixed in 552dc12 — `Search` now refetches the token and retries the POST once, returning `ErrCSRFRequired` only on a second 403; `TestSearchRefreshesTokenOnForbidden` expects the stale-token call to succeed and `TestSearchPersistentForbidden` pins the give-up path.

internal/ats/dayforce.go
amikai: Fixed in 552dc12 — both regexes capture the culture, it is persisted on the minted slug (`{ns}/{board}/{culture}`, omitted when it is the default) and carried onto the resolved `Company` for SiteInfo, search, and detail; `en-US` remains only for locale-less URLs and roster rows without `culture_code`. 91e8312 additionally rejects the two-segment URLs where the culture was being minted as the namespace.

The locale-less half of your last sentence is now the only gap left, tracked in a separate follow-up comment.

internal/ats/dayforce.go
amikai: Fixed in 552dc12, tidied in 91e8312 — `distanceUnit` goes out with every `distance` in both the adapter and the CLI (which now also sends `distance`), behind a named `distanceUnitMiles` constant, and `search_location_req.hurl` records the payload. `locationId`/`locationType` are still omitted: they are zeros in the captured body and the endpoint narrowed correctly without them.

internal/provider/dayforce/siteinfo.go
amikai: Fixed in 552dc12 — `ClientId` is gone, `parseSiteInfo` rejects `jobBoardId <= 0`, and `TestSiteInfoMissingJobBoardID` / `TestSiteInfoAllowsNullClientID` pin both halves. 91e8312 also removes the blocking part of your note: Search no longer fetches SiteInfo at all.

internal/provider/dayforce/mocksrv.go
amikai: Fixed in 552dc12 — all three spots you named are trimmed: `MockBoolAttributeJobID` is down to the constraint, `BoardClient`'s doc comment is two lines with the naming history dropped, and the 403 comment now reads "Refresh once and retry; a second 403 remains a CSRF error", which matches what the code does.

internal/provider/dayforce/openapi.yaml
amikai: Partly done in 552dc12 — the risky ones (typed non-null scalars: `jobReqId`, `jobDescription`, `searchScore`, `isEvergreen`, `reapplyPeriodDays`, `postingExpiryTimestampUTC`, `isInternal`, `relocationEligible`, `created`/`lastModifiedTimestampUTC`, and the location sub-fields) are no longer `required`.

Still `required` but read by nothing outside tests: `SearchPosting.clientNamespace`, `SearchPosting.jobBoardId`, `JobPostingDetail.isoCurrencyRegion`, `JobPostingDetail.jobPostingAttributes`. Leaving this open since it is your thread — a tenant omitting any of the four still costs a whole page or detail call, which is the rule the spec's own note states.

amikai (commented):
Re-reviewed at c8caf73. Both remaining threads are addressed and resolved. Re-ran the paging probe to be sure the empty-branch change did not disturb it: 352 jobs over 18 pages of 20 (12 on the last), contiguous, no duplicates, empty past the end, and zero SSR requests on the roster path. `go build`, `go vet`, `go test ./...` (with `-race` on the provider), `make validate-openapi`, `make hurl-lint` all clean; the spec relaxation was regenerated rather than hand-edited, and the four now-optional fields have no non-test readers.

One nit on the probe guard below, which is a flaw in the snippet I handed you.

internal/ats/dayforce.go
```diff
@@ -0,0 +1,587 @@
...
+	if start >= first.MaxCount {
+		// A non-roster board's culture is assumed, and an unsupported culture
+		// returns 200/maxCount 0 rather than an error — so an empty result is
+		// the one place worth paying for SiteInfo, which 404s on a culture the
+		// board does not serve.
+		if board.JobBoardID == 0 {
+			if _, err := a.jobBoardID(ctx, &board); err != nil {
+				return nil, err
+			}
+		}
```
amikai, line 360: **Nit, and my fault — the guard I suggested is one condition too broad.** `start >= first.MaxCount` is also true for a page past the end of a *non-empty* board, so an out-of-range page on a non-roster board pays the probe (and inherits its failure) even though `maxCount` already answered the question.

Verified against a synthetic non-roster board with `maxCount: 352` and an SSR page that 404s:

```
page 1  (in range): err=<nil> jobs=20
page 99 (past end): ERROR dayforce: resolve board otherns/otherboard: fetch dayforce site-info page: unexpected status 404
```

Page 1 works, page 99 hard-errors — but nothing about the board is wrong, the caller just asked past the end. The trap you are guarding is specifically `maxCount == 0`:

```suggestion
	if start >= first.MaxCount {
		// A non-roster board's culture is assumed, and an unsupported culture
		// returns 200/maxCount 0 rather than an error — so a board that reports
		// no jobs at all is the one place worth paying for SiteInfo, which 404s
		// on a culture the board does not serve. A page merely past the end of a
		// non-empty board needs no probe.
		if first.MaxCount == 0 && board.JobBoardID == 0 {
			if _, err := a.jobBoardID(ctx, &board); err != nil {
				return nil, err
			}
		}
```

internal/ats/dayforce.go
amikai: Fixed in c8caf73 — the probe now runs on the zero-result path only, and `TestDayforceSearchUnsupportedCulture` / `TestDayforceSearchEmptyNonRosterBoard` pin both outcomes (404 surfaces, genuinely empty board stays empty). Verified the happy path is untouched: a roster search makes no SSR request at all, and the probe is memoized after the first miss.

One leftover in the guard's condition — my snippet's flaw, not yours — is in a separate comment.

internal/provider/dayforce/openapi.yaml
amikai: Fully done in c8caf73 — `SearchPosting.clientNamespace`/`jobBoardId` and `JobPostingDetail.isoCurrencyRegion`/`jobPostingAttributes` are no longer `required`, so nothing that the adapter never reads can fail a page or a detail call any more. Regenerated (schemas, decoders, validators) rather than hand-edited, and `go test ./...` is green with the `Opt`-wrapped assertions.

internal/ats/dayforce.go
amikai: Fixed in 81a9430 — the probe is now gated on `first.MaxCount == 0`, and `TestDayforceSearchPastEndNonRosterSkipsSiteInfo` pins it with a mock tenant (`nossr`) that serves a full search page while its SSR page 404s. I checked that test against the previous guard and it fails there with the exact error from the report, so the regression is genuinely covered rather than just fixed.

Full suite, `-race` on the provider, `make validate-openapi` and `make hurl-lint` all green at 81a9430. Nothing further from me — this closes every thread on the PR.

internal/ats/dayforce.go
```diff
@@ -0,0 +1,588 @@
...
+
+	req := dayforce.SearchRequest{
+		ClientNamespace: board.Namespace,
+		JobBoardCode:    board.JobBoardCode,
+		CultureCode:     board.Culture(),
+	}
+	if p.Query != "" {
+		req.SearchText = dayforce.NewOptString(p.Query)
+	}
+	if location := strings.TrimSpace(p.Location); location != "" {
```
amikai: **[bug] `location="remote"` is sent to the geo search and returns false negatives.** The unified MCP schema explicitly promises that `location: "remote"` matches remote-friendly jobs, but this branch sends it as a geocoded `locationString` with a 50-mile radius. At this HEAD, a live PCA request returned `maxCount: 0`; walking its 14 unfiltered pages found 8 postings with `hasVirtualLocation: true` (including fixture job 62239). Please special-case `remote`: use an upstream virtual-work filter if one exists, or filter candidate pages on `HasVirtualLocation` before applying unified pagination, and add a regression test.

internal/ats/dayforce.go
```diff
@@ -0,0 +1,588 @@
...
+	payClasses, err := a.client.PayClasses(ctx, board.Namespace, board.JobBoardID, board.Culture())
+	if err != nil {
+		return nil, fmt.Errorf("dayforce: filters %q: %w", slug, err)
+	}
+	payTypes, err := a.client.PayTypes(ctx, board.Namespace, board.JobBoardID, board.Culture())
+	if err != nil {
+		return nil, fmt.Errorf("dayforce: filters %q: %w", slug, err)
+	}
+
+	return FilterSet{
```
amikai, line 595: **[bug] `travel_required` is accepted by `applySearchFilters` but cannot be discovered.** The company-tool schema tells clients that filter keys and values are company-specific and must come from `get_filters_by_company`; this result exposes only three dimensions, so a conforming client will never send the fourth supported key. Please include a static `travel_required: ["true", "false"]` dimension (similar to SmartRecruiters' static `location_type` values).

internal/provider/dayforce/mocksrv.go
```diff
@@ -0,0 +1,236 @@
...
+	var req SearchRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return mockSearchRsp
+	}
+	if text, ok := req.SearchText.Get(); ok && text == "electrical engineer" {
+		return mockSearchFilteredRsp
+	}
+	// Location branch pins locationString/distance/distanceUnit; there is
+	// no live-captured location page, so reuse the filtered fixture.
+	if _, ok := req.LocationString.Get(); ok {
```
amikai: **[test gap] This branch does not actually pin `distance` or `distanceUnit`.** Any request containing `locationString` receives the filtered fixture, so `TestDayforceSearchByLocation` would still pass if the adapter regressed to the earlier payload that omitted `distanceUnit`. Decode and verify the expected location, `distance: 50`, and `distanceUnit: 0`, failing the mock request (or returning the base fixture) when any differ. The Hurl fixture is static and does not exercise the Go request builder.

amikai (commented):
Three inline findings from the latest review.

PR #301, opened 2026-08-14
feat(adp_wfn): serve ADP Workforce Now career centers

### LINKED ISSUE #300: adp_wfn: integrate ADP Workforce Now career centers
## Problem Statement

ADP runs two unrelated public career-board surfaces. We serve one of them —
ADP MyJobs — and the other, ADP Workforce Now, is invisible to every client of
this server. A caller who names a company hosted on Workforce Now, or who
pastes its careers URL, gets "unknown company" and no path forward. The
comments in the ADP MyJobs provider and adapter have named `adp_wfn` as a
future package since that provider was built; this closes it.

Workforce Now is not a niche corner of ADP. Its boards are ordinary employers
publishing ordinary openings, reachable over a public JSON API with no auth,
and they are currently unreachable through the unified company tools purely
because nobody has written the adapter.

## Solution

A new `adp_wfn` provider and its `ats.Adapter`, so Workforce Now companies
answer the unified company tools exactly like every other ATS-backed company.
A caller names a company, or pastes a Workforce Now careers URL, and gets
paged job listings, the filter dimensions that company's board actually
publishes, and full job detail — without ever learning which platform served
it.

The provider follows the established pipeline: a trimmed OpenAPI spec and a
generated client, a fixture-replaying mock server, a debug CLI, the adapter,
and registration in the server. Per ADR 0001 it is a separate package from ADP
MyJobs; the two share a vendor and nothing else.

### Finding jobs

1. As a job seeker, I want to search a Workforce Now company by name, so that
   I do not need to know which hiring platform that employer uses.
2. As a job seeker, I want to search a Workforce Now company by pasting its
   careers URL, so that I can reach an employer that is not in the curated
   roster.
3. As a job seeker, I want to page through a company's openings, so that I can
   see the whole board rather than only the first screen.
4. As a job seeker, I want the reported total count to be honest, so that I do
   not conclude a board is larger or smaller than it is.
5. As a job seeker, I want a keyword search across a company's postings, so
   that I can narrow a large board to the roles I care about.
6. As a job seeker, I want keyword search to match description text and not
   only titles, so that a role described in the body but titled generically
   still surfaces.
7. As a job seeker, I want a quoted phrase in my keyword search to be treated
   as a phrase, so that I can search precisely when a loose match returns too
   much.
8. As a job seeker, I want to be told plainly that keyword search returns one
   page of relevance-ranked results, so that I do not assume I am paging an
   exhaustive list.
9. As a job seeker, I want to filter a company's openings by location, so that
   I only see roles I could actually take.
10. As a job seeker, I want a location I name to be matched against the
    dimension values that company's board publishes, so that I am not silently
    handed the whole unfiltered board.
11. As a job seeker, I want to filter by the job types a company publishes
    (full-time, part-time, per-diem, and whatever else that tenant configured),
    so that I can exclude arrangements I cannot accept.
12. As a job seeker, I want to combine a keyword, a location, and a job type in
    one search, so that I can narrow on several axes at once.
13. As a job seeker, I want to pass several values for one filter dimension and
    get the union, so that I can search three cities in one request.
14. As a job seeker, I want location and job-type filtering to page soundly, so
    that I can walk a filtered result set without missing or repeating rows.

### Understanding a posting

15. As a job seeker, I want the full description of a posting, so that I can
    decide whether to apply without leaving my client.
16. As a job seeker, I want the description as plain text, so that it is
    readable rather than raw markup.
17. As a job seeker, I want pay information when the employer published it, so
    that I can filter on compensation before investing time.
18. As a job seeker, I want a clickable public URL for each posting, so that I
    can open and apply on the employer's own page.
19. As a job seeker, I want that URL to open a working page rather than an
    empty one, so that a link is never a dead end.
20. As a job seeker, I want the posting's location and posted date on the
    summary, so that I can triage a list without opening every row.
21. As a job seeker, I want the company name on a job detail, so that I know
    who I would be applying to.

### Discovering what a board supports

22. As a job seeker, I want to ask which filter dimensions a company's board
    offers, so that I can filter without guessing at value names.
23. As a job seeker, I want those dimensions reported per company, so that a
    board filing jobs by store and a board filing them by state each describe
    themselves honestly.
24. As a job seeker, I want an unknown filter key or value to be rejected with
    the available options, so that a typo produces guidance rather than wrong
    results.

### Reaching companies not on the roster

25. As a job seeker, I want a current-form Workforce Now careers URL to be
    accepted, so that I can search an employer nobody has curated yet.
26. As a job seeker, I want a legacy Workforce Now careers URL to be accepted
    too, so that an old link I found still works.
27. As a job seeker, I want a legacy link that cannot be resolved to fail
    quickly with a message telling me the current URL form, so that I am not
    left guessing.
28. As a job seeker, I want a job URL from an unlisted company to still be
    usable as company input, so that I can go from one posting back to that
    employer's whole board.
29. As a job seeker searching an unlisted company, I want the company name that
    the platform itself records, so that the result is not labelled with an
    opaque identifier.

### Correctness a user should never have to think about

30. As a job seeker, I want a company whose board is published in a non-US
    locale to return its jobs, so that a Canadian employer does not appear
    empty.
31. As a job seeker, I want an empty result to mean the company has no matching
    openings, so that I can trust emptiness rather than suspect a
    misconfiguration.
32. As a job seeker, I want a filter I did not ask for never to be silently
    dropped, so that the rows I see are the rows I asked for.
33. As a job seeker, I want the first page of results to start at the first
    job, so that no posting is skipped.

### Curating the roster

34. As a roster curator, I want each roster entry to carry everything needed to
    address its tenant, so that a company resolves without runtime discovery.
35. As a roster curator, I want a candidate company that returns no jobs to
    fail verification, so that a misconfigured entry cannot enter the roster
    looking healthy.
36. As a roster curator, I want to look up a tenant's recorded company name
    from its identifier, so that I can name a roster entry from the platform
    rather than from guesswork.
37. As a roster curator, I want to resolve a readable company slug to its
    tenant identifier, so that bulk discovery has a path that scales.
38. As a roster curator, I want roster entries keyed by a readable slug, so
    that a caller who mistypes a company name still gets a useful suggestion.
39. As a roster curator, I want the seed roster to be small and provably
    correct, so that the integration is reviewable and bulk growth stays a
    separate change.
40. As a roster curator, I want an entry's name cross-checked against more than
    one source before it lands, so that a plausible-but-wrong name cannot ship.

### Maintaining the provider

41. As a maintainer, I want a debug CLI with search, detail, and companies
    subcommands, so that I can check the live surface without going through the
    server.
42. As a maintainer, I want each known upstream trap captured as a fixture, so
    that a change in upstream behavior fails a test rather than silently
    corrupting results.
43. As a maintainer, I want every quirk of the surface documented where the
    code lives, so that the next person does not rediscover it by hand.
44. As a maintainer, I want the provider registered in the server and smoke
    tested end to end, so that "done" means reachable rather than merely
    written.
45. As a maintainer, I want this provider to leave the ADP MyJobs provider
    untouched, so that its own hard-won quirks cannot regress.
46. As a maintainer, I want the two ADP comments that call `adp_wfn` a future
    package updated, so that the codebase stops describing something that now
    exists.

### Package boundary

Per ADR 0001, a new provider package and a new adapter, entirely separate from
ADP MyJobs. Companies appearing on both surfaces are expected; cross-adapter
name collisions are resolved by careers URL at runtime and only need naming in
the collision-report golden file.

### Surface and client

Public JSON REST with no auth, no token, and no cookies on the data endpoints.
The provider gets a hand-written OpenAPI spec trimmed to the endpoints used —
job list, job detail, search filters, tenant features, and career-center
content links — and a generated client. The spec is the ground truth for the
discovered quirks; operation descriptions carry them so they survive
regeneration.

The tenant is addressed by a GUID. Every other identifier is rejected by the
API, so the roster stores the GUID even though callers never see it.

### Search shape: server-side paging, not a full dump

The upstream paginates natively with a page cap equal to this repo's uniform
page size, so pagination passes straight through. The provider is **not** a
full dump and must not synthesize one.

Page offsets are one-based. The page-size parameter is honored only on
unfiltered requests and ignored entirely under any filter, so paging must
always step by a full page. Termination is by empty result array, never by a
computed page total.

### Filter model

Query-parameter filters are silently ignored by the upstream — an unrecognized
parameter returns the whole board with a success status. The working controls
are one query parameter for keyword relevance and two request headers for
location and job type.

Both header filters have a silent-fallback hazard, and they differ in shape.
An unpaired location value returns the whole unfiltered board, while a
well-formed location naming nothing returns an empty set. A job-type value the
tenant does not publish does something worse, corrected during implementation
after recon reported it as a whole-board fallback: it returns a large arbitrary
**subset** — on a 21-posting board, three unrelated bogus values each returned
the identical 19 rows with an inflated total. A whole-board result is
recognisable as unfiltered; a 19-of-21 result is not. **Every value must
therefore be validated before it is sent.** Locations require a
value/qualifier pair; job types require an identifier drawn from the tenant's
own published filter dimensions.

The tenant's published dimensions invert which column goes on the wire between
the two: locations send the published value, job types send the published
identifier. Sending the wrong column returns the whole board.

A tenant may publish no location dimension and still accept location
filtering. Absence of a published dimension means there is no list to validate
against, not that the capability is missing.

Filter values use OR semantics within a dimension and AND across dimensions,
matching the documented meaning of the unified filter type. The keyword,
location, and job-type controls compose as a true AND upstream.

Filter dimensions are fetched per company on demand. There is no cache: the
cost of omitting one is a constant extra request rather than a growing one,
and adding caching later is a pure optimization that changes no interface.

### Keyword search is page one only

Upstream relevance ordering is nondeterministic between identical calls, so
consecutive pages are drawn from differently ordered candidate lists and
overlap. Keyword search therefore serves a single page of relevance-ranked
results and reports one total page. Paging it with client-side de-duplication
would be silent incompleteness, which this codebase does not do. Location and
job-type filtering page soundly and are the tool for completeness.

### Result counts

The upstream total is trustworthy only on an unfiltered request; under any
filter it becomes a relevance tally that routinely exceeds the board. Filtered
searches therefore report the lower bound the walk proved, incremented when a
further page exists so that the page total still signals more — the same
pattern the Avature adapter already uses for portals whose counts are
unknowable from one page.

### Locale

A tenant's jobs exist in exactly one locale; every other locale that tenant
advertises returns an empty set, so a single locale per roster entry loses
nothing and a list would invite a pointless fan-out.

A recognized but wrong locale returns an empty set, and an absent or
unrecognized locale falls back to a server default. Both failure modes are
indistinguishable from a company with no openings, so the locale is a
**required** roster field validated at load time, not an optional field with a
default.

For a company reached by careers URL, the locale is read from the URL when it
carries one. When it does not, the career-center content-links resource is
consulted, whose first advertised locale is the job-bearing one. The tenant
features resource also reports a default locale, but it is wrong for at least
one verified tenant and must not be used for this.

### Company identity

Roster entries carry a curated readable slug plus the tenant GUID, the
platform's own client identifier, the career-center identifier, and the
locale. The slug is curated rather than taken from the platform because the
registry builds its typo-suggestion pool from slugs only and never from names:
a GUID slug would make every company in this provider unsuggestible and would
print a GUID in an error telling a human to type a slug. The platform's own
identifier is opaque for a meaningful share of tenants, so it is stored for
discovery rather than used as the key.

For a company reached by careers URL and not on the roster, the company name
comes from the platform's tenant features resource and is passed through
verbatim. Its casing and truncation are imperfect; correcting them would be
manufacturing a name rather than reporting one. The job list and job detail
resources carry no company name at all.

### Careers URL handling

The current URL form is parsed directly. The legacy form carries a readable
slug and resolves to the tenant GUID through a redirect chain that requires a
cookie handshake — without one it redirects to itself indefinitely, which is
why it appears dead. That resolution happens inside URL parsing with a
three-second timeout and a temporary cookie-bearing client that shares the
existing transport; the cookie jar must not be attached to the shared client
or cookies leak into every other provider's requests. Resolution does not
succeed for every slug, and a failure returns "not a careers URL for this ATS"
so the caller is guided to the current form.

Two consequences are accepted deliberately: the parse is not cancellable by
the client, because the adapter interface takes no context and changing it
would touch every adapter; and the worst case adds the timeout to an input
form the vendor has already decommissioned.

For a roster company the parse yields that company's curated slug. For an
unlisted tenant it yields a canonical careers URL as the slug, so that the
tenant identifier, career-center identifier, and locale all survive into
search and detail. Without this the locale on the input URL is lost the moment
parsing returns, and a company whose board is not in the default locale
silently reports zero jobs. This mirrors what the Workday, UltiPro, and Oracle
adapters already do.

### Job identity and URLs

The external job identifier is present on every row observed and is used as
the job id, because it is also the key the public posting URL requires. The
internal item identifier also works for detail lookups but cannot build a URL.
A third identifier present on rows is a decoy: it returns a hollow record
rather than an error.

Rendered URLs use the company's own locale rather than a fixed default, so a
link never opens an empty board.

### Detail

Detail adds exactly one field over the list row — the description — which is
the only reason to make the call. Pay and level fields are per-row optional
and identical between list and detail. Pay is surfaced as a leading line in
the plain-text description, since the unified job detail type has no
structured salary field; adding one is follow-up work spanning every adapter.

### Career-center identifier

Omitted from data requests: it is not the constant it first appears to be, a
wrong value returns an empty set rather than an error, and omitting it worked
on every tenant tested. It is retained in the roster for rendering
human-facing URLs, since a tenant publishes it in its own public link and the
page's behavior without it has not been verified.

### Registration

The adapter is registered in the server's ATS registry, its careers-host
pattern is added to the registry's pattern map so the unrecognized-URL error
advertises it, and it is added to the verification command's provider order.
The integration is complete only when sampled roster companies return live
listings through the real server path and a job detail resolves for a returned
job id.

### What makes a good test here

Tests assert external behavior through the adapter's own interface: given a
slug and search parameters, the returned jobs, counts, and page numbers. They
do not assert which headers were composed, which query parameters were built,
or how many upstream calls were made — those are the implementation choices
most likely to change. The exception is the small set of cases where the
upstream's silent failures are only observable as a difference in outcome; for
those, the fixture encodes the upstream's behavior and the test asserts our
result, not our request.

### Seams

The primary seam is **the adapter interface against the provider's
fixture-replaying mock server**, which is the pattern every existing ATS
adapter already uses: the mock server replays captured fixtures, and the test
points the adapter's base-URL fields at it. One seam covers roster loading,
careers-URL parsing, careers-URL rendering, search, filter discovery, and
detail.

A second, lower seam already established by convention is the generated client
against the same mock server, exercising request construction and response
decoding.

One new seam point is required: the legacy careers-URL resolver's host is
currently implicit, and URL parsing now performs network I/O. That host
becomes a field on the adapter struct, matching the existing base-URL field
pattern, so tests can point it at the mock server. This is the highest
available point — no new interface, no injection layer.

The live fixture replay against the real site remains a separate check for
fixture rot, not a unit test.

### Prior art

The ADP MyJobs adapter tests are the closest model: roster-builds-registry,
a table-driven careers-URL parse test including negative cases from
neighbouring platforms, and search tests driven by the provider's mock server.
The Dayforce provider is the model for fixture layout — many small request and
response pairs rather than one large capture.

### What gets tested

- Roster loads and builds a registry without collisions; every entry renders a
  careers URL.
- Careers-URL parsing: current form, current form carrying a job id, legacy
  form that resolves, legacy form that does not resolve, neighbouring ADP
  hosts, and unrelated hosts.
- An unlisted tenant parses to a slug that carries its locale, and searching
  with that slug reaches the right board.
- Unfiltered search paging, including that the first page starts at the first
  job and that a page boundary does not repeat or skip rows.
- Filtered search paging terminates on an empty result rather than on a
  computed total.
- Reported totals: passthrough when unfiltered, lower bound when filtered.
- Keyword search returns a single page and reports one total page.
- A filter value that cannot be validated is rejected with the available
  options, and is never forwarded.
- A tenant publishing no location dimension still accepts location filtering.
- A company whose board is in a non-default locale returns its jobs; the same
  company under the default locale returns none.
- Detail returns plain text, carries pay when the row has it, and omits it
  cleanly when the row does not.
- An unlisted tenant's detail carries the platform-recorded company name; a
  roster company's detail carries the roster name.

### Fixtures to capture

Beyond the happy path, detail, not-found, and unknown-tenant captures, each
known trap is captured as its own request/response pair so that a change in
upstream behavior fails loudly: the unpaired-location-token whole-board
fallback; a well-formed location pair with values that match nothing; the
one-based versus zero-based page offset; the inflated total under a filter;
the non-default-locale board returning empty under the default locale; and a
tenant publishing no location dimension. Where a live count is volatile, tests
assert structure rather than the exact number.

## Out of Scope

- **Bulk roster growth.** The seed roster proves the pipeline; expanding it is
  a separate session and a separate roster-prefixed change, per the repository
  commit convention.
- **ADP MyJobs.** Not modified, beyond updating the two comments that describe
  `adp_wfn` as a future package.
- **The ADP developer and partner APIs**, and the internal employee career
  center, which explicitly refuses external candidates.
- **A structured salary field on the unified job detail type.** WFN publishes
  structured pay and the unified type cannot hold it; adding one touches every
  adapter and belongs in its own change.
- **Teaching the registry's suggestion pool to rank company names as well as
  slugs.** This is a latent gap for every adapter, not just this one, and
  should not ride along here.
- **Optical recognition of tenant logos.** It is the only source that names
  every tenant, but it belongs to roster curation, not to a running server,
  and the server will not carry a vision dependency to fill an optional field.
- **Caching.** Deliberately omitted; see the implementation decisions.
- **Sorting.** The upstream sort controls have no observable effect.

## Further Notes

Several conclusions in this specification reverse an earlier reading of the
same surface, and the reversals are worth knowing because each one hides a
trap:

- The job-type header filter appears inert on a tenant that has exactly one
  job type, because filtering by it returns the whole board.
- Pay and level fields appear to be list-only if the compared detail record
  happens to be one that lacks them.
- The legacy careers URL appears dead because without a cookie jar it
  redirects to itself.
- The list of locales a tenant advertises is the set its career-center UI is
  translated into, not the set that has postings.
- The tenant features resource reports a default locale that is wrong for at
  least one verified tenant.
- A resource carrying tenant identity sits outside the versioned path prefix,
  so an endpoint sweep of that prefix misses it.

During research, one tenant was confidently misidentified from a social-media
link, and the wrong answer was plausible rather than obviously broken. That is
the reason roster names are cross-checked against more than one independent
source before an entry lands, and the reason the platform's own recorded name
is passed through verbatim rather than reconstructed from slugs, domains, or
logo filenames.

The vendor publishes no crawl policy on this host: the robots file redirects
to a login page for both a browser and a search-engine agent, so there is
neither a directive to honor nor a permission to cite. The data is
unauthenticated public job postings, no bot detection or rate limiting was
observed, and request pacing stays polite.

Full design record, including every option considered and the measurements
behind these decisions, is in the dated `adp_wfn` provider plan under the
repository's plans directory. The package-boundary decision is recorded as
ADR 0001. Vocabulary used here follows the repository glossary.

amikai:
Closes #300.

Adds `adp_wfn` as an ATS adapter, so Workforce Now companies answer the unified company tools. Kept separate from `adp_myjobs` per `docs/adr/0001` — same vendor, otherwise unrelated surfaces.

Almost nothing on this API reports a problem with a status code, so the design is mostly about telling wrong answers apart from empty ones:

- filter values are validated against the tenant's own catalog before they are sent — an unpaired location returns the whole board, and an unpublished job-type oid returns a 19-of-21 subset that looks filtered
- counts are recomputed whenever a filter is applied, since the upstream's own count becomes a relevance tally that overshoots the board
- keyword search serves one page: upstream relevance ordering is not stable between identical calls, so paging it overlaps and drops rows
- roster `locale` is required, not defaulted — the wrong one returns an empty board rather than an error

Verified live: 16/16 hurl fixtures, `verify-companies` green on all 6 roster companies, and an MCP stdio smoke test covering search by name, an `en_CA` board, filters, careers-URL input, and job detail.

Design record in `docs/superpowers/plans/2026-08-14-adp_wfn-provider.md`.


internal/ats/adp_wfn.go
```diff
@@ -0,0 +1,479 @@
...
+// The upstream answers an unknown locale by falling back to en_US, and answers
+// a known-but-wrong one with an empty board. Both are indistinguishable from a
+// tenant with no openings, so guessing is not an option when the tenant did
+// not tell us.
+func (a *ADPWFNAdapter) locale(ctx context.Context, t adpWFNTenant) string {
+	if t.locale != "" {
+		return t.locale
+	}
+	locale, err := a.client().PrimaryLocale(ctx, t.cid)
+	if err != nil || locale == "" {
```
amikai: **[bug]** `locale()` swallows `PrimaryLocale` errors and returns `""`, after which `Search`/`Filters`/`Detail` proceed without a locale. Omitting locale is the upstream's en_US fallback, which this PR itself documents as indistinguishable from an empty board. Roster rows always carry a locale, so the MCP happy path is fine — but an unlisted tenant reached by a URL without `lang=` (including every resolved `posting.html?client=` slug) will report zero openings on any content-links timeout, 5xx, or decode error, instead of failing the call. That is exactly the silent-total-loss case the adapter comments say must not happen.

**Suggestion:** Change `locale()` to return `(string, error)` and propagate a `PrimaryLocale` failure from `Search`, `Filters`, and `Detail`. Returning `""` only when the tenant advertises no locale is a separate, documented choice; a transport/decode error is not.

cmd/adp_wfn/main.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	params := adp_wfn.ListParams{Locale: tenantLocale, Query: strings.TrimSpace(keyword), Page: page}
+	if location = strings.TrimSpace(location); location != "" {
+		if !strings.Contains(location, ",") {
+			return fmt.Errorf("--location %q forms no value,qualifier pair; upstream would ignore it and return the whole board. "+
+				"Use %q or %q", location, location+",LOCATION_CITY", "IN,LOCATION_STATE")
+		}
+		params.Locations = []string{location}
+	}
+	if jobType = strings.TrimSpace(jobType); jobType != "" {
```
amikai, line 236: **[suggestion]** The debug CLI forwards `--job-type` straight onto `workerCategoriesList` with no catalog check. That is the worse of the two silent fallbacks: an unpublished oid or a copied display label returns a large plausible subset (19 of 21 in the captured fixture), not an error and not the whole board. `--location` at least refuses an unpaired token; `--job-type` has no equivalent guard. The `filters` output prints `label -> oid`, so pasting the left column is an easy mistake.

**Suggestion:** Resolve the tenant's `SearchFilters` catalog and accept only a published oid or label, the same way `ADPWFNAdapter.resolveFilters` does. Reject anything else before calling `List`.

internal/provider/adp_wfn/client.go
```diff
@@ -0,0 +1,491 @@
...
+	Locale string
+	// Query is a relevance probe over the description text, and only the
+	// first page of its results is meaningful; see [BoardClient.List].
+	Query string
+	// Locations must already be `<value>,<qualifier>` pairs. Sending a bare
+	// token returns the whole unfiltered board, so callers validate before
+	// they reach here.
+	Locations []string
+	// WorkerCategories must be oids published by [BoardClient.SearchFilters].
+	// A label, or an oid from another tenant, returns the whole board.
```
amikai: **[suggestion]** `ListParams.WorkerCategories` says a label or a foreign oid "returns the whole board". That is the first-pass recon that the plan and `TestListBogusJobTypeReturnsMisleadingSubset` later corrected: the live answer is a large arbitrary subset with an inflated total. Callers who trust this comment may treat a bogus oid as "obviously unfiltered" when it looks filtered.

**Suggestion:** Replace the sentence with the verified behavior (unpublished oid → misleading subset, HTTP 200) and keep the "validate against `SearchFilters` first" requirement.

internal/ats/adp_wfn.go
```diff
@@ -0,0 +1,479 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	jobs := make([]JobSummary, 0, len(res.Jobs))
+	for _, r := range res.Jobs {
+		id := adp_wfn.ExternalJobID(r)
+		if id == "" {
+			id = strings.TrimSpace(r.ItemID.Value)
```
amikai, line 231: **[suggestion]** When `ExternalJobID` is missing, Search/Detail fall back to `itemID` and then pass that id to `JobURL`. `JobURL`'s own godoc says an itemID produces a share link that does not resolve. The plan claimed 100% ExternalJobID coverage and "no fallback needed"; a fallback that emits a dead URL is worse than skipping the row.

**Suggestion:** If `ExternalJobID` is empty, skip the list row (Search already does this when both ids are empty) and, in Detail, keep `itemID` only as the `JobID` used to fetch, not as the `jobId` query value. Or omit `URL` rather than publish a broken permalink.

internal/ats/adp_wfn_test.go
```diff
@@ -0,0 +1,310 @@
...
+	a := newADPWFNAdapter(t)
+	ctx := context.Background()
+
+	_, err := a.Search(ctx, adp_wfn.MockSlug, SearchParams{Filters: FilterSet{"department": []string{"x"}}})
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "unknown filter key")
+
+	// A value that is not published must never reach the wire: upstream would
+	// answer it with the whole board rather than an error.
+	_, err = a.Search(ctx, adp_wfn.MockSlug, SearchParams{Filters: FilterSet{adpWFNLocationKey: []string{"Atlantis"}}})
```
amikai, line 192: **[suggestion]** `TestADPWFNSearchRejectsUnvalidatableFilters` covers an unknown key and an unpublished location, and the nearby comment says the danger is "the whole board". The hazard that actually looks like a successful filter is an unpublished job-type oid (`9:999` → 19 of 21). Nothing in the adapter tests asserts that `job_type=9:999` (or a label) is rejected and never put on the wire.

**Suggestion:** Add a case that `Filters: {job_type: [MockBogusJobTypeOID]}` (and a published label sent as the value, which should map to the oid) errors or, for the label, sends only the catalog oid. The client-level "bogus oid returns 19 rows" test should stay as documentation of why the adapter must not reach that path.

internal/provider/adp_wfn/companies.go
```diff
@@ -0,0 +1,87 @@
...
+
+// mustLoadCompanies parses the embedded companies.yaml. A parse failure is a
+// build-time bug in a file this package owns, not a runtime condition to
+// recover from.
+func mustLoadCompanies() []Company {
+	var cs []Company
+	if err := yaml.Unmarshal(companiesYAML, &cs); err != nil {
+		panic(fmt.Sprintf("adp_wfn: parse companies.yaml: %v", err))
+	}
+	for _, c := range cs {
```
amikai, line 71: **[suggestion]** `mustLoadCompanies` checks required fields but not duplicate `slug` or `cid`. `buildSlugIndex` / `buildCIDIndex` last-write-wins, so a curation mistake would silently drop a tenant from URL resolution (`CompaniesByCID` is what `ParseCareersURL` uses). Sibling `adp_myjobs` rejects duplicate slugs/names at load; `NewRegistry` only catches same-adapter slug/name collisions, not two slugs sharing a cid.

**Suggestion:** Panic (or return an error from a testable loader) on duplicate lowercase slug or cid, matching `adp_myjobs.loadCompanies`. The existing `TestCompaniesRosterIsWellFormed` check is not a load-time guarantee.

cmd/verify-companies/main.go
```diff
@@ -236,6 +239,13 @@ func buildAdapters(names []string) ([]ats.Adapter, error) {
 		if err != nil {
 			return nil, fmt.Errorf("build %s adapter: %w", name, err)
 		}
+		// A provider listed in providerOrder with no case above leaves a nil
```
amikai: **[suggestion]** The new nil-adapter guard is good, but the comment narrates the incident that produced it ("that failure lands on every invocation…") rather than stating the invariant.

**Suggestion:** Keep the `a == nil` check; shrink the comment to why a nil adapter is possible (a `providerOrder` name with no switch case).

cmd/adp_wfn/main.go
```diff
@@ -0,0 +1,330 @@
...
+				return fmt.Errorf("companies takes no positional arguments, got %v", args)
+			}
+			return runCompanies(*format)
+		},
+	})
+
+	searchFS := ff.NewFlagSet("search").SetParent(rootFlags)
+	var (
+		keyword  = searchFS.StringLong("keyword", "", "relevance search over the description; first page only")
+		location = searchFS.StringLong("location", "", "one published location value, or a city or two-letter state")
```
amikai: **[nit]** `--location` help says a two-letter state is accepted, but `runSearch` rejects any value without a comma, so `--location IN` fails even though the adapter would turn it into `IN,LOCATION_STATE`.

**Suggestion:** Either construct the same pairs `adpWFNLocationPair` does, or change the help/error to require an already-paired value (`City, ST` / `IN,LOCATION_STATE`).

amikai (commented):
## Summary

This is a complete, well-researched ADP Workforce Now ATS integration: ogen client, seed roster, mock+hurl fixtures for the silent-fallback traps, debug CLI, adapter, and MCP/`verify-companies` wiring all land together. Pagination (`$skip` 1-based), locale pinning, header filters, and 404-HTML handling match the captured surface, and the provider is actually registered. The remaining issues are a real silent-empty-board path when locale discovery fails, plus a few safety-comment/CLI gaps around the unpublished job-type oid hazard.

## Issue counts by severity

- bugs: 1
- suggestions: 6
- nits: 1

amikai (commented):
Two additional findings from the ADP WFN provider review.

cmd/adp_wfn/main.go
```diff
@@ -0,0 +1,330 @@
...
+func runSearch(ctx context.Context, company, cid, locale, keyword, location, jobType string, page int, format string) error {
+	tenantCID, tenantCCID, tenantLocale, err := resolveTenant(ctx, company, cid, locale)
+	if err != nil {
+		return err
+	}
+	client, err := newClient()
+	if err != nil {
+		return err
+	}
+	params := adp_wfn.ListParams{Locale: tenantLocale, Query: strings.TrimSpace(keyword), Page: page}
```
amikai: **[bug]** The CLI leaves `Page` caller-controlled when `--keyword` is set, even though `BoardClient.List` explicitly says keyword results must not be paged. Recon found that later windows overlap and omit jobs because relevance ordering changes between identical calls, so `--keyword welder --page 2` can return a confidently incomplete result. Reject pages above 1 when a keyword is present, or force page 1 before constructing `ListParams`.

internal/provider/adp_wfn/oas_parameters_gen.go
```diff
@@ -0,0 +1,865 @@
...
+	// empty string, punctuation — answers the whole unfiltered board with HTTP 200. Callers must
+	// therefore validate that every value forms a pair before sending it, and must never forward caller
+	// input unchecked.
+	//
+	// Feed values from `getSearchFilters` verbatim: that operation publishes them already pair-formed.
+	LocationsList OptString `json:",omitempty,omitzero"`
+	// Job-type filter, taking the `oid` values published by `getSearchFilters`. Multiple comma-separated
+	// oids OR together.
+	//
+	// Carries the same silent-fallback hazard as `locationsList`: a bogus oid, or the human-readable label
```
amikai: **[maintenance]** The checked-in generated comment is stale relative to `openapi.yaml`. Running `go generate ./internal/provider/adp_wfn` immediately modifies this block, so the PR is not generation-clean and a future regeneration will introduce an unrelated diff. Regenerate and commit the ogen output; the current uncommitted workspace change contains the exact generated delta.

internal/ats/adp_wfn.go
amikai: Took the alternative you offered at the end rather than the primary suggestion, so flagging the difference.

Skipping the row loses a real posting: `Detail` accepts `itemID` as well, so a row without an `ExternalJobID` is still fetchable and still worth returning — it just cannot be linked. Dropping it would hide a job that the adapter can otherwise serve in full.

So both `Search` and `Detail` now split the two uses of the id: `JobID` falls back to `itemID` and stays addressable, while `URL` is left empty unless the external id is present. Nothing emits a permalink it knows will 404.

Worth noting the fallback is still unexercised in practice — `ExternalJobID` was on 100% of rows across both tenants swept — so this is about what happens the day that stops being true, not a case anyone hits today.

internal/provider/adp_wfn/client.go
```diff
@@ -0,0 +1,495 @@
...
+//
+// Row offsets are 1-based upstream: asking for offset 0 silently drops a row.
+// When p carries a Query, only page 1 is meaningful — the upstream orders
+// relevance results nondeterministically between identical calls, so
+// consecutive windows overlap and drop rows. Callers must not page a query.
+func (c *BoardClient) List(ctx context.Context, cid string, p ListParams) (*ListResult, error) {
+	page := max(p.Page, 1)
+	params := ListJobRequisitionsParams{
+		Cid:  cid,
+		Skip: NewOptInt32(int32((page-1)*PageSize + 1)),
```
amikai, line 141: **[bug]** `page` is user-controlled and the MCP schema has no maximum, but this arithmetic is converted to `int32` without a bounds check. For example, `page = math.MaxInt` overflows the `int` calculation to `-39`, so the client sends a negative `$skip`; an upstream that clamps or ignores it can return page 1 while the adapter labels the result as the enormous requested page. Guard the largest page whose 1-based offset fits in `int32` before calculating it, matching the overflow checks in the Avature and Workday adapters.

PR #302, opened 2026-08-16
feat(foxconn): model the filter-code endpoints, retiring the frozen copy

amikai:
## Summary

Models the two `Labels/*/Codes` reference endpoints on Foxconn's Taiwan careers API and retires the hand-transcribed copy of the same enums, plus smaller fixes to the Synopsys and MediaTek clients and to the server's flag handling.

This PR **does not** wire MediaTek, Foxconn, Quanta, Realtek or Synopsys into the MCP server. Their client packages and debug CLIs land here; exposing them as tools is a separate decision.

## Motivation & Context

Foxconn's board serves its `workplaceCode` and `talentZoneCode` enums from two reference endpoints. The spec had declared them deliberately out of scope while `codes.go` carried a hand-transcribed copy, so the transcription was the only place anyone could read them from. Modelling them leaves one source of truth — the board's own.

The copy still matched both endpoints exactly when checked (2026-08-16), so nothing had drifted. Talent zones are keyed by dated recruitment campaigns, which is what makes a frozen copy a question of when rather than if.

Verifying that turned up two surface quirks, now recorded in the spec:

- **`workplaceCode` is not ISO 3166.** It looks like alpha-2 and three values do coincide, but `CH` is China, `AM` is the USA, `BZ` is Brazil, and `ID`/`IND` are inverted against ISO — India/Indonesia here, Indonesia/India there. An unknown code returns HTTP 200 with an empty array rather than a 404, so a guessed ISO code reads as "no jobs in that country" instead of failing.
- **zh-TW only.** None of `Accept-Language`, `?lang=`, `?locale=`, `?language=`, `?culture=` or a `lang` cookie changes the response, and no sibling English portal exists. Only job titles are mixed, because employers type some in English.

## Key Changes

**Foxconn** — `listWorkplaceCodes` / `listTalentZoneCodes` added to the spec and generated client; `codes.go` removed and the `foxconn codes` subcommand reads the endpoints instead; the two deviations above documented; `.hurl` captures added for the new fixtures.

**Synopsys** — `NewClient` takes a `Config` so the base URL is injectable; the posting URL is built from the package's own `orgID` rather than a second hardcoded copy; the package has one mock server instead of two.

**MediaTek** — the job ID regex accepts subsidiary prefixes, not only `MTK`.

**openings-mcp** — flags bind to `OPENINGS_MCP_*` through `ff.WithEnvVarPrefix`, so an explicit flag beats the environment and an unparseable value fails at startup instead of silently falling back to the default. `--http` now rejects an empty value, which an exported-but-empty `OPENINGS_MCP_HTTP` would otherwise turn into a listener bound to no address. Also restores the assertions covering every registered tool's description, schemas and annotations, which nothing else was checking.

## Notes for review

- Roughly three quarters of the diff is regenerated ogen output; the hand-written portion is a few hundred lines.
- `ResolveLocation`, `LocationSuggestion` and `AsFilter` are exported on the Synopsys client but have no caller outside the package yet. They are the surface an MCP integration would need.

## Verification

- `go test -race -vet=all ./...`, `make validate-openapi`, `make hurl-lint` and `gofmt` all clean.
- The two new `.hurl` files run green against the live endpoints, and both committed fixtures match the live responses exactly.
- `foxconn codes` exercised against the live board.


internal/openingsmcp/foxconn.go
```diff
@@ -0,0 +1,245 @@
...
+var foxconnSearchInputRawSchema = []byte(`{
+	"type": "object",
+	"properties": {
+		"keyword": {
+			"type": "string",
+			"description": "Case-insensitive free-text search across job title and description."
+		},
+		"workplace": {
+			"type": "string",
+			"description": "Workplace location filter code. These are NOT ISO 3166 country codes even though they look like them: TA is Taiwan (not TW), CH is China (not CN), VM is Vietnam (not VN), AM is the USA (not US), and ID/IND are India/Indonesia — the opposite way round from ISO. A code the board does not know returns zero results instead of an error, so a guessed ISO code reads as 'no jobs in that country'. Call foxconn_get_search_filters for the real codes rather than guessing."
```
amikai:
```suggestion
			"description": "Workplace location filter code. Call foxconn_get_search_filters for the real codes rather than guessing."
```

internal/openingsmcp/foxconn.go
```diff
@@ -0,0 +1,245 @@
...
+		},
+		"talent_zone": {
+			"type": "string",
+			"description": "Talent zone / recruitment track filter code (e.g. TALENTS for general recruitment, MA for management associate, IAI_KS for Kaohsiung software center). Call foxconn_get_search_filters to see all available codes."
```
amikai:
```suggestion
			"description": "Talent zone / recruitment track filter code. Call foxconn_get_search_filters to see all available codes."
```

PR #315, opened 2026-08-17
feat(freehire): add job-board MCP tools

### LINKED ISSUE #314: Add freehire.me as a source — 227 boards behind one key-free API
## Add freehire.me as a source

openings-mcp already fans out across boards and company career sites from one MCP client. freehire fits that model exactly — except it is itself an aggregation of 227 sources, so one backend adds coverage that would otherwise be many adapters.

It is also an MCP server in its own right (`me.freehire/freehire` in the official registry), so if openings-mcp ever wants to delegate rather than adapt, that path exists too.

[freehire.me](https://freehire.me) is an open-source IT job aggregator that crawls postings straight from company career boards and ATS platforms — 227 sources, 3.3M+ open roles across 294K companies — normalized into one schema and deduplicated. Live figures: [/open](https://freehire.me/open).

**`GET /api/v1/jobs/search` answers 200 signed out** — no key, no cookie, no approval queue.

What a source adapter gets:

- Postings pre-tagged with stack, seniority, region, work mode, employment type and English level, plus a server-side visa/sponsorship filter
- Full descriptions as markdown **inside the search response**, so scoring needs no per-hit detail fetch
- The company's own ATS `url` on every result — the apply link stays the employer's, not an aggregator's
- Filter vocabulary discoverable at runtime via `GET /api/v1/jobs/facets` (canonical skill slugs, country codes, enums with live counts), so nothing is hardcoded against a vocabulary that changes

Links: OpenAPI https://freehire.me/openapi.yaml · docs https://freehire.me/docs/api · backend (Go, MIT) https://github.com/strelov1/freehire · CLI, `--json` on every command https://github.com/strelov1/freehire-cli

Scope note: IT/tech roles only, so it complements existing sources rather than replacing them.

Disclosure: I maintain freehire. Happy to write the adapter and open a PR if you want it — asking first rather than sending unsolicited code.

amikai:
## Summary

Adds [freehire.me](https://freehire.me) as a dedicated job-board source: four MCP tools over an ingest-time catalogue of IT/tech postings crawled from company ATS boards, with keyword, facet, salary, and company filtering that all runs server-side.

Closes #314.

## Motivation & Context

freehire is a catalogue over many ATS boards, not a tenant ATS. Company is a search facet (`company_slug`), not a tenant, so putting it behind the unified company tools would collide with employers this server already serves first-hand — Stripe has 576 openings on Greenhouse directly and 520 via freehire — and would answer with the stale subset. Dedicated tools keep company as one optional filter among many.

What freehire adds that no existing provider does is the enrichment layer: normalized skill/seniority/category slugs across ~1.5M postings, and a `reality` block carrying freehire's own freshness verdict (`class`, `age_days`, `repost_count`, `fake_freshness`).

## Architecture

```mermaid
flowchart LR
  Client[MCP client]
  Client -->|"name → company_slug"| Companies["freehire_search_companies<br/>GET /companies"]
  Client -->|"uncertain filter value"| Facets["freehire_get_job_facets<br/>GET /jobs/facets"]
  Companies --> Search
  Facets --> Search
  Client --> Search["freehire_search_jobs<br/>GET /agent/jobs/search"]
  Search -->|public_slug| Detail["freehire_get_job_detail<br/>GET /jobs/{slug}"]
```

Every filter needs somewhere to get a legal value from, because an invalid one returns HTTP 200 with `total: 0` rather than an error. The schema enums `seniority`, `work_mode`, `regions`, `sort`, and `order`; `freehire_get_job_facets` resolves the open-vocabulary ones; `freehire_search_companies` resolves `company`, which the facets response has no facet for.

## Key Changes

- **Client**: ogen from the official `openapi.yaml`, used unmodified. `window.go` records `MaxResultWindow`, one of two limits the published spec omits.
- **Search**: keyword, company, skills, seniority, work mode, regions, country, source, category, salary range, semantic ratio, sort/order, paging. Rows carry `company_slug`, `regions`, `countries`, `posted_at`, `created_at`, `closed_at`, `enrichment`, `reality`, and a truncated plain-text description opening.
- **Facets**: a `facets` selector and `max_values` cap, because the full vocabulary is ~86 KB across two dozen facets.
- **Companies**: fuzzy name → `company_slug` resolution, returning a list because one name can belong to several employers.
- **CLI**: `cmd/freehire` with `search`, `facets`, `companies`, `detail`.
- **Server**: four tools registered in `newServer`; server instructions tell hosts to URL-dedupe freehire against `search_jobs_by_company`.

## Reviewer Guidance

Read `docs/superpowers/specs/2026-08-17-freehire-provider-design.md`, then `internal/openingsmcp/freehire.go` (the whole MCP surface), then `internal/provider/freehire/testdata/*.hurl`. The `oas_*_gen.go` files are ogen output — 11k of the 16.5k added lines.

Worth a closer look: the per-row `description` is truncated to 400 runes (`freehireDescriptionChars`) because the search endpoint returns each row's full posting regardless of `description_format` and offers no way to opt out, so a 20-row page is ~25 KB of tool output either way. That constant is the main tunable if the balance looks wrong.

## Verification & Testing

- `go test ./...` — 73 packages, plus `go vet ./...` and `gofmt`.
- All 9 hurl fixtures replay against the live API (`make hurl-test`), including `jobs_deep_page_req.hurl`, which pins the undocumented paging ceiling: `offset+limit` may reach 10000, past which the API answers `400 {"error":"pagination too deep"}`.
- Live MCP over stdio: fuzzy company resolution (`altinty` → `altinity-inc`) feeding a company-filtered search; `salary_min=100000` narrowing 2868 hits to 213; `sort=posted_at&order=desc` returning same-day postings; facet selection cutting a 54 KB response to 910 bytes; page 501 rejected at the schema.
- Live CLI: all four subcommands.

## Risks & Rollout

`semantic_ratio` is exposed because the spec declares it, but ratio 0 and ratio 1 returned identical rankings on every query tried — it may be inert upstream.

Two parameters work live but are absent from the published spec and therefore not exposed: `role` and `visa_sponsorship`. Adding them would mean maintaining an annotated `openapi.yaml`; the deliberate choice here is to stay on the upstream spec verbatim.

Detail responses are freehire's crawled snapshot, which can be stale — the detail fixture is a posting freehire itself marks `class: stale` after two reposts. The employer apply URL is the authoritative source, and the tool description says so.


cmd/openings-mcp/main.go
```diff
@@ -66,11 +67,11 @@ var (
...
-- When the user explicitly names a job board or careers site as the desired source (for example LinkedIn, Indeed, 104, Cake.me, Jobindex, マイナビ転職/Mynavi, Flowxtra, Amazon Jobs, Apple Careers, Google Careers, Meta Careers, or TSMC Careers), use that source's dedicated tools. A company name by itself is not a source selection.
+- When the user names a specific company, try search_jobs_by_company first; it covers thousands of companies and its error message suggests close matches when a name isn't recognized. If it does not recognize the company, try freehire second: freehire_search_companies turns an approximate or misspelled name into the company_slug that freehire_search_jobs takes, and freehire crawls many ATS platforms this server has no adapter for. Fall back to the per-provider tools (linkedin, indeed, 104, jobindex, mynavi, ...) only after both, since they search by keyword rather than by company.
```
amikai: **[P1] Prefer first-party company tools before freehire**

This ordering sends any company missed by `search_jobs_by_company` to freehire before all dedicated providers. For example, a request for Google jobs can succeed against the crawled freehire snapshot and never reach `google_search_jobs`, despite the direct first-party source being available. Route dedicated company-career tools before freehire, and reserve freehire for companies without a first-party integration.

internal/openingsmcp/freehire.go
```diff
@@ -0,0 +1,1230 @@
...
+		if in.CompanySlug == "" {
+			return errorResult(fmt.Errorf("company_slug is required (take it from a freehire_search_companies result)")), nil, nil
+		}
+		res, err := c.GetCompany(ctx, freehire.GetCompanyParams{
+			Slug:   in.CompanySlug,
+			Limit:  freehireOptInt(in.Limit),
+			Offset: freehireOptInt(in.Offset),
+		})
+		if err != nil {
+			return errorResult(fmt.Errorf("company %q not found", in.CompanySlug)), nil, nil
```
amikai: **[P2] Preserve non-404 detail errors**

Every client error is discarded and reported as not found, so timeouts, invalid responses, rate limits, and upstream failures look like stale slugs and lose their diagnostic cause. Use `errors.As` to inspect `freehire.ErrorStatusCode`, translate only status 404, and wrap or return the original error otherwise. The job-detail handler repeats the same issue below.

docs/superpowers/specs/2026-08-17-freehire-provider-design.md
```diff
@@ -0,0 +1,38 @@
...
+Dedicated job-board tools, not an `ats.Adapter`.
+
+freehire is an ingest-time catalogue (Meilisearch over ~3.3M IT postings)
+with a keyless JSON API. Company search is a facet (`company_slug`), not
+a tenant. Stripe on freehire is a crawled Greenhouse subset of a company
+already on the Greenhouse roster, so putting it behind
+`search_jobs_by_company` would collide and return an incomplete board.
+
+MCP surface: `freehire_search_jobs` + `freehire_get_job_facets` +
+`freehire_search_companies` + `freehire_get_job_detail`.
```
amikai, line 12: **[P2] Update the design to match v1.1.0**

This newly added design still documents the first implementation, while the final commit registers six tools, wires company detail and city search, uses `/jobs/search` previews, accepts `countries`, and exposes `visa_sponsorship`. Because the reviewer guidance and plan link here as the architectural contract, update this document and the accompanying plan to describe the shipped surface.

amikai (commented):
Inline review findings

cmd/openings-mcp/main.go
```diff
@@ -66,11 +67,11 @@ var (
...
+- When the user names a specific company, prefer the most direct source for it, in this order. (1) If that company has its own tools here — amazon, apple, google, meta, tsmc — use them: they read the employer's own careers site, and none of these five is in search_jobs_by_company's roster. (2) Otherwise try search_jobs_by_company; it covers thousands of companies and its error message suggests close matches when a name isn't recognized. (3) Only if neither covers the company, try freehire: freehire_search_companies turns an approximate or misspelled name into the company_slug that freehire_search_jobs takes, and freehire crawls many ATS platforms this server has no adapter for. freehire holds a crawled snapshot of an employer's board, so never let it stand in for a first-party source that exists. (4) Fall back to the keyword boards (linkedin, indeed, 104, jobindex, mynavi, ...) last, since they search by keyword rather than by company.
```
amikai, line 73: **[P2] Do not claim these companies are absent from the unified roster**

The new first-party-first order is correct, but the supporting claim is false in the current tree: the unified roster has exact `Amazon` and `Apple` entries in Workable, `Meta` in SmartRecruiters, and `TSMC` in Avature. Those entries may be collisions or inferior substitutes, which is an even better reason to prefer the dedicated tools, but telling the host that none exists makes the routing contract disagree with what `search_jobs_by_company` can resolve. Remove this sentence or rephrase it to say that unified matches are not authoritative substitutes for the dedicated first-party tools.

Issue #303, opened 2026-08-16
feat(mcp): support dynamic filter discovery and relax rigid schema enums across remaining providers


amikai:
## Background & Motivation

Following recent improvements to **Foxconn**, **Realtek**, **MediaTek**, and **Quanta** (along with existing filter tools in **Apple**, **Meta**, and the unified `get_filters_by_company`), several remaining standalone job board and enterprise career site providers still rely on rigid static `enum` arrays in their MCP tool `InputSchema` and/or lack dedicated `*_get_search_filters` discovery tools.

### Why this is a problem
1. **Schema Fragility**: If an upstream company or job board updates its taxonomy (adds new office locations, reorganizes job categories, or introduces new subsidiary entities), queries using the updated terms fail client-side MCP `InputSchema` validation before even reaching the API.
2. **LLM Discoverability**: Without a `*_get_search_filters` tool, LLM agents have to guess opaque codes or exact uppercase string constants (e.g., `USA-Arizona` vs `Arizona`, or `aws` vs `AWS`).

---

### 1. TSMC (`tsmc`) — High Priority
- **Current State**:
  - `tsmc_search_jobs` hardcodes static `enum`s for `location` (17 items, e.g. `Taiwan`, `USA-Arizona`, `Japan-Kumamoto`), `category` (21 items, e.g. `R&D`, `IC Design Technology`), `job_type` (5 items), and `employment_type` (4 items).
  - No `tsmc_get_search_filters` tool exists.
- **Proposed Solution**:
  - Add `tsmc_get_search_filters` tool returning all valid locations, categories, job types, and employment types.
  - Relax `tsmc_search_jobs` `InputSchema` from rigid `enum`s to open `string` fields, with backend mapping/case-insensitive matching against TSMC facet codes.

### 2. NVIDIA (`nvidia`) — High Priority
- **Current State**:
  - `nvidia_search_jobs` hardcodes static `enum`s for `job_category` (14 items), `job_type` (6 items), `country` (35 countries), `location_type` (`Office`/`Remote`), and `time_type`.
  - No `nvidia_get_search_filters` tool exists.
  - Upstream is Workday (`nvidia.wd5.myworkdayjobs.com`), which natively supports dynamic facet discovery.
- **Proposed Solution**:
  - Add `nvidia_get_search_filters` tool (querying Workday facets or returning the taxonomy).
  - Relax `nvidia_search_jobs` `InputSchema` from strict `enum`s to flexible `string`s.

[...]

PR #317, opened 2026-08-18
feat(cake): add search filter discovery and relax schema enums

amikai:
## Summary
Adds the `cake_get_search_filters` MCP tool for dynamic facet discovery and relaxes strict schema enums across Cake.me search tools with alias and label normalization.

## Motivation & Context
Resolves the Cake provider portion of #303.
Previously, Cake.me job search schema enforced strict enums for `job_type`, `seniority`, and `remote`. Queries with alternative casing or human-readable names (e.g., `"全職"`, `"中高階"`, `"純遠端"`, `"Full-time"`) failed client-side validation, and LLM callers lacked a way to inspect valid filter dimensions.

## Key Changes
- **OpenAPI Schema & Models**:
  - Modeled `AvailableFacets` on `JobSearchResponse` in `internal/provider/cake/openapi.yaml`.
  - Relaxed `JobSearchFilters` strict enum constraints to open string slices with `additionalProperties: true`.
  - Regenerated ogen client code via `go generate ./internal/provider/cake/...`.
- **MCP Server Tools (`internal/openingsmcp`)**:
  - Registered `cake_get_search_filters` tool to query live facet options (`job_types`, `seniority_levels`, `remote`, `year_of_seniority`, `locations`, `professions`) and enrich them with human-readable Chinese/English labels.
  - Relaxed `cakeSearchInputRawSchema` input schema.
  - Added input normalization for `job_type`, `seniority`, `remote`, and `sort` parameters to support both localized labels and raw slugs.
- **CLI & Tests**:
  - Updated `cmd/cake` CLI and test suite for relaxed filter types.
  - Added unit and E2E tests for `cake_get_search_filters` and label normalization.

## Verification & Testing
- Provider tests: `go test -v ./internal/provider/cake/...`
- MCP tests: `go test -v ./internal/openingsmcp/... -run TestCake`
- Repository test suite: `go test ./...`


internal/openingsmcp/cake.go
```diff
@@ -187,6 +292,63 @@ func cakeHTTPToMCPResponse(resp *cake.JobSearchResponse) *cakeSearchOutput {
...
+		}
+		return out
+	}
+
+	for _, item := range facets.JobTypes {
+		name := item
+		if label, exists := cakeJobTypeLabels[item]; exists {
+			name = label
+		}
+		out.JobTypes = append(out.JobTypes, cakeFilterOption{Value: item, Name: name})
```
amikai, line 393: The discovery output does not round-trip through the new label normalization. For example, this returns `Name: "全職 / Full-time"`, but `normalizeCakeJobType` lowercases that to `"全職 / full-time"` and does not match any case, so it is sent upstream as an unknown slug. Cake silently ignores unknown filter values, which makes the search appear successful while returning unfiltered results. Please either make every returned `Name` normalize back to its `Value`, or explicitly make the contract value-only instead of advertising label inputs.

PR #318, opened 2026-08-20
feat(freehire): sync upstream openapi.yaml and set project User-Agent

amikai:
## Summary
Synchronizes the local `internal/provider/freehire/openapi.yaml` with the latest upstream specification from freehire.me and adds a dedicated HTTP transport setting a project-identifying `User-Agent`.

## Motivation & Context
Upstream freehire.me updated its public OpenAPI specification with:
1. Documented rate limiting guidelines (600 rpm for standard reads, 300 rpm for `agentSearchJobs`).
2. An upstream request for callers to provide an identifying `User-Agent` (`owner/project` format) so upstream maintainers can reach out before applying breaking changes or hard limits.

## Key Changes
- **OpenAPI Specification**: Updated `internal/provider/freehire/openapi.yaml` to match upstream.
- **Transport & User-Agent**: Added `freehire.Transport` in `internal/provider/freehire/transport.go` with default `User-Agent: amikai/openings-mcp (+https://github.com/amikai/openings-mcp)`.
- **MCP Server & CLI**: Wired `freehire.Transport` into `cmd/openings-mcp/main.go` and `cmd/freehire/main.go`.
- **Documentation**: Updated `internal/provider/freehire/doc.go` documenting rate limits and identification behavior.

## Verification & Testing
- Unit tests added in `internal/provider/freehire/transport_test.go` covering default and custom `User-Agent` values.
- All test suites passing (`go test ./...`).
- `go vet ./...` executed cleanly with no issues.


amikai (commented):
Review of the User-Agent wiring and the spec sync.

Verified separately, so these are *not* problems: `internal/provider/freehire/openapi.yaml` is byte-identical to the live `https://freehire.me/openapi.yaml` (still 1.1.0), `go generate ./internal/provider/freehire/` produces no drift, and `go test -race -vet=all ./...` is clean.

Eight inline notes below, plus one that has no diff line to hang on:

**The `.hurl` fixtures still hit freehire.me anonymously.** `make hurl-test` fires all 15 `internal/provider/freehire/testdata/*.hurl` at the live API with hurl's own default User-Agent — so the traffic that actually bursts against upstream is the traffic that stays unidentified, while the doc now claims we identify ourselves. `internal/provider/job104/testdata/job_detail_req.hurl:4` mirrors its transport's headers for exactly this reason.

internal/provider/freehire/transport.go
```diff
@@ -0,0 +1,29 @@
...
+// DefaultUserAgent names this project and points to its public repository,
+// matching the identification request in freehire's openapi.yaml.
+const DefaultUserAgent = "amikai/openings-mcp (+https://github.com/amikai/openings-mcp)"
+
+// Transport sets the User-Agent header on outgoing requests to freehire.me.
+// When UserAgent is empty, [DefaultUserAgent] is used. Wrap it around an
+// *http.Client and pass that to NewClient via WithClient.
+type Transport struct {
+	Base      http.RoundTripper
+	UserAgent string
```
amikai: `UserAgent` has no production caller — only `transport_test.go` sets it, so as shipped this is configuration with no consumer.

The caller that would make it earn its place is the one that also fixes what the UA is missing: `version` in `cmd/openings-mcp/main.go` is set at release time by `.goreleaser.yaml` (`-X main.version=...`) and is already in scope at the construction site, so

```go
freehire.Transport{UserAgent: "amikai/openings-mcp/" + version + " (+https://github.com/amikai/openings-mcp)"}
```

is one line. The spec's own example carries a version (`acme/job-sync/1.4 (+…)`), and the stated payoff — a heads-up before a limit or a field changes — is precisely the case where upstream needs to tell a months-old build from a current one. Right now every build of every user's binary sends the same undifferentiated string.

internal/provider/freehire/transport.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func (t Transport) RoundTrip(req *http.Request) (*http.Response, error) {
```
amikai, line 15: Third verbatim copy of this body: `internal/provider/job104/transport.go` and `internal/provider/eightfold/transport.go` declare the same struct shape and the same `clone → set headers → nil-Base fallback` skeleton, differing only in which headers get set.

Each copy re-implements the `Clone` contract and the fallback independently, so fixing either is a three-place edit. A shared `httpx.HeaderTransport{Base http.RoundTripper; Header http.Header}` carries it once — otherwise the next provider that needs headers makes four.

internal/provider/freehire/transport.go
```diff
@@ -0,0 +1,29 @@
...
+	req = req.Clone(req.Context())
+	ua := t.UserAgent
+	if ua == "" {
+		ua = DefaultUserAgent
+	}
+	req.Header.Set("User-Agent", ua)
+	base := t.Base
```
amikai, line 18: This branch is never exercised — both tests leave `Base` nil. Rewrite `RoundTrip` to call `http.DefaultTransport` unconditionally and the suite still passes: it only asserts the header the handler observed over a real socket, which `DefaultTransport` delivers just as well. A caller stacking `Transport{Base: instrumentedRT{}}` would lose their transport silently.

Same gap on line 18: nothing asserts the caller's original request is left unmodified, which is the only reason the `Clone` is there.

internal/provider/freehire/transport_test.go
```diff
@@ -0,0 +1,49 @@
...
+func TestTransportDefaultUserAgent(t *testing.T) {
+	var gotUA string
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		gotUA = r.Header.Get("User-Agent")
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer srv.Close()
+
+	client := &http.Client{Transport: Transport{}}
+	req, err := http.NewRequest(http.MethodGet, srv.URL, nil)
```
amikai: Context-less `http.NewRequest` is a first for this repo's tests — 825 `t.Context()` call sites, zero other `http.NewRequest(` in `_test.go`. A request on the background context isn't cancelled when the test times out, so a hung handler blocks until the whole `go test` binary timeout fires instead of failing this one test. Line 41 too.

Lines 18 and 37: this package's own `client_test.go` uses `t.Cleanup(srv.Close)` rather than `defer`.

internal/provider/freehire/transport_test.go
```diff
@@ -0,0 +1,49 @@
...
+	require.NoError(t, err)
+
+	resp, err := client.Do(req)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+
+	assert.Equal(t, DefaultUserAgent, gotUA)
+}
+
+func TestTransportCustomUserAgent(t *testing.T) {
```
amikai: Verbatim copy of `TestTransportDefaultUserAgent` apart from the `Transport` literal and the expected string. Adding the cases that are currently missing (`Base` delegation, request-not-mutated) means a third and fourth copy.

A table over `{name, transport, wantUA}` sharing one capture handler covers all of them — and the coverage gap then reads as a missing row instead of a missing function.

internal/provider/freehire/doc.go
```diff
@@ -13,6 +13,12 @@
 //     picked.
 //   - getSimilarJobs is not wired into the MCP server yet.
 //
+// # Rate limits and identification
```
amikai: This documents the 429 / `Retry-After` / `X-RateLimit-*` contract without anything reading it. `openapi.yaml` declares only `"200"` and a `default` per path, so a 429 arrives at `internal/openingsmcp/freehire.go:938` as a plain `*freehire.ErrorStatusCode`, where only 404 is special-cased. The MCP host then sees an opaque upstream error carrying no "retry in N seconds", and the natural LLM response — retry immediately — spends the rest of the window.

Nothing loops over pages today, so it's latent rather than live. Worth either handling 429 or saying here that it isn't handled.

internal/provider/freehire/doc.go
```diff
@@ -13,6 +13,12 @@
...
+//
+// freehire enforces per-caller rate limits (600 rpm across standard reads,
```
amikai: These two figures already live in `openapi.yaml`, in the same package, in a file the README describes as synced verbatim from upstream. When upstream retunes the budgets, the next sync updates the yaml and leaves this line asserting the old numbers — with no fixture pinning them, unlike every other divergence the "Where the spec and the live API disagree" section tracks.

Pointing at the spec's "Rate limits" section keeps one copy.

cmd/openings-mcp/main.go
```diff
@@ -338,7 +338,8 @@ func newProviderServer(logger *slog.Logger, dumpCache *ats.DumpCache) (*mcp.Serv
 		return nil, fmt.Errorf("create Flowxtra client: %w", err)
 	}
 
-	cFreehire, err := freehire.NewClient("https://freehire.me/api/v1", freehire.WithClient(hc))
+	hcFreehire := &http.Client{Timeout: 30 * time.Second, Transport: freehire.Transport{}}
```
amikai, line 343: Third client to opt out of the shared `hc`, and the only one that doesn't say why. Line 310 explains the shared pool and its ceiling; line 313 explains Eightfold's 403 on Go's default User-Agent.

A reader here has to open `internal/provider/freehire/transport.go` to learn that freehire's exception is a voluntary courtesy header rather than a hard requirement — and someone consolidating clients later can fold it back into `hc` with nothing here or in any test failing.

internal/provider/freehire/transport.go
amikai: Fixed — `UserAgent(version)` at `transport.go:16`, wired at `cmd/openings-mcp/main.go:344`, with a table covering both the empty and the unexpanded-placeholder cases. Resolving. One follow-up on that function itself in the new review.

internal/provider/freehire/transport.go
amikai: Fixed — extracted to `internal/httpx.HeaderTransport` with all three providers delegating to it. Resolving the duplication itself; the extraction did lose two things on the way, raised as new comments on `internal/httpx/transport.go:17` and `internal/provider/job104/transport.go:23`.

internal/provider/freehire/transport.go
amikai: Fixed — `captureRoundTripper` covers the `Base` branch in both `internal/httpx/transport_test.go` and this package's test, and `assert.Empty(t, req.Header.Get("User-Agent"))` now pins the Clone contract. Resolving.

internal/provider/freehire/transport_test.go
amikai: Fixed — `http.NewRequestWithContext(t.Context(), ...)` and `t.Cleanup` throughout, in both test files. Resolving.

internal/provider/freehire/transport_test.go
amikai: Fixed — table-driven now, and the missing coverage did land as new rows rather than new functions. Resolving.

internal/provider/freehire/doc.go
amikai: Leaving this one open, deliberately.

Deleting the section does remove the inconsistency I flagged — the doc no longer claims an awareness the code doesn't have. But neither remedy happened: a 429 still reaches `internal/openingsmcp/freehire.go:938` as an opaque `*freehire.ErrorStatusCode` sitting right next to the 404 special case, and now nothing anywhere records that it's unhandled.

Closing it as "openapi.yaml is the only place this gets written down" is a perfectly reasonable call — it's just yours to make, not mine, so I'm not resolving it for you.

internal/provider/freehire/doc.go
amikai: Fixed — section removed, so `openapi.yaml` stays the single copy of the figures. Resolving.

cmd/openings-mcp/main.go
amikai: Fixed — comment added at `main.go:341`. Resolving.

internal/httpx/transport.go
```diff
@@ -0,0 +1,24 @@
...
+	Base   http.RoundTripper
+	Header http.Header
+}
+
+// RoundTrip clones the request, applies all headers in Header, and delegates
+// to Base (or http.DefaultTransport if Base is nil).
+func (t HeaderTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	cloned := req.Clone(req.Context())
+	for k, vv := range t.Header {
+		cloned.Header[k] = vv
```
amikai: `cloned.Header[k] = vv` bypasses `http.Header`'s key canonicalization, which the `Header.Set` calls in the three bodies this replaces were doing for free. A non-canonical key is stored verbatim, `net/http` then fails to find it when it decides whether to write its own defaults, and both go out. Confirmed against this branch:

```go
HeaderTransport{Header: http.Header{"user-agent": {"mine/1.0"}}}
// server receives:
//   User-Agent: []string{"Go-http-client/1.1", "mine/1.0"}
```

Harmless for today's three callers, which all pass canonical literals. But this is now a shared `internal/` API, and the failure mode is exactly the one it exists to prevent — leaking `Go-http-client/1.1` to the edge you were trying to look ordinary to, which is precisely what Eightfold 403s on.

Worth cloning the slice while you're there: the assignment aliases the transport's own long-lived `Header` value into every request it forwards, so a later `Add` on a request whose slice has spare capacity can write back into it.

```go
for k, vv := range t.Header {
	cloned.Header[http.CanonicalHeaderKey(k)] = slices.Clone(vv)
}
```

internal/provider/job104/transport.go
```diff
@@ -10,16 +14,13 @@ type BrowserTransport struct {
...
-		base = http.DefaultTransport
-	}
-	return base.RoundTrip(req)
+	return httpx.HeaderTransport{
+		Base: t.Base,
+		Header: http.Header{
+			"User-Agent":      []string{"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36"},
+			"Accept":          []string{"application/json, text/plain, */*"},
+			"Accept-Language": []string{"zh-TW,zh;q=0.9,en-US;q=0.8,en;q=0.7"},
+			"Referer":         []string{"https://www.104.com.tw/"},
```
amikai: The extraction dropped the note that used to sit on this line:

> A same-origin Referer is required — Cloudflare 403s otherwise; a generic one (not the exact source page) is accepted for every path.

It's gone from the repo entirely — `git grep "same-origin Referer"` is empty at this head. Of the four headers here it's the one whose value looks arbitrary and isn't, and the "a generic one works for every path" half is what someone will otherwise rediscover by collecting 403s.

My reuse suggestion caused this, so: worth carrying back above the `Referer` entry.

internal/provider/freehire/transport.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// DefaultUserAgent names this project and points to its public repository,
+// matching the identification request in freehire's openapi.yaml.
+const DefaultUserAgent = "amikai/openings-mcp (+https://github.com/amikai/openings-mcp)"
+
+// UserAgent returns the formatted User-Agent for a given version string.
+// If version is empty or "version" (the default unexpanded build placeholder),
+// [DefaultUserAgent] is returned.
+func UserAgent(version string) string {
```
amikai: Two things about this living here.

`version == "version"` encodes a `cmd/openings-mcp` build detail — the unexpanded `main.version` placeholder — inside a provider package that otherwise has no relationship with the binary. From in here it reads as a magic string. The caller is the one that knows what its own placeholder means, so either `UserAgent` takes a version it can trust, or the placeholder check sits next to the `var version = "version"` it's actually about.

Second, `"amikai/openings-mcp"` and the repo URL are now written twice — once in `DefaultUserAgent`, once in the format string — so a rename has to hit both. `DefaultUserAgent = UserAgent("")` would collapse that, or both could compose one project/contact pair.

Neither is load-bearing: the behavior is right and the table test covers it.

internal/provider/eightfold/transport.go
```diff
@@ -11,12 +15,11 @@ type BrowserTransport struct {
...
-	req.Header.Set("User-Agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36")
-	req.Header.Set("Accept", "application/json, text/plain, */*")
-	base := t.Base
-	if base == nil {
-		base = http.DefaultTransport
-	}
-	return base.RoundTrip(req)
+	return httpx.HeaderTransport{
+		Base: t.Base,
+		Header: http.Header{
```
amikai: This `http.Header` literal — and job104's four-entry one — is now rebuilt on every request: a map plus one slice per header, per call. The contents are constant, so a package-level `var browserHeaders = http.Header{...}` reads the same and allocates once.

Negligible against the round trip it precedes; raising it only because it's free while these lines are already being touched.

amikai (commented):
Re-review of the force-pushed head (`0acc810`).

Seven of the eight threads are fixed and resolved. The `doc.go` 429 one I replied to and left open — deleting the section removes the inconsistency but takes neither of the two remedies, and which way to close it is your call.

The `.hurl` gap from the last review body is fixed too: all 20 requests across the 15 fixtures now carry the identifying UA.

`go build ./...`, `go vet ./...` and `go test -race -vet=all` are clean on `internal/httpx`, all three transports and `internal/openingsmcp`.

Four new notes, all fallout from the `httpx` extraction. One is a real (currently dormant) behavior change; the rest are cosmetic.

internal/provider/freehire/doc.go
amikai: Your call — closing this. Checking the actual surface, the decision holds up better than my comment implied:

- `ErrorStatusCode.Error()` renders as `code 429: {...}`, so the status and upstream's message both reach the host verbatim. The LLM has enough to recognize a throttle and back off.
- `freehireDetailError`'s own comment already records the intent: *"reporting a timeout, a rate limit, or an unparseable body as 'not found' would send the caller hunting for a stale slug instead of retrying."* So the "nothing records that it's unhandled" half of my earlier reply was wrong — it is written down, just in `internal/openingsmcp/freehire.go` rather than `doc.go`.

The one thing that doesn't reach the host is `Retry-After`, since it's a header rather than part of the body: the LLM learns it was throttled, not for how long. Reasonable to accept.

Resolving.

internal/provider/freehire/transport.go
amikai: Head is back at `25d9e40`, so this is live again — reopening.

Given version isn't going in yet, the answer changes: instead of wiring one through, drop the `UserAgent` field and set `DefaultUserAgent` directly. That makes `Transport` exactly the shape of its two siblings (`eightfold.BrowserTransport`, `job104.BrowserTransport`), removes a knob nothing sets, and takes `TestTransportCustomUserAgent` with it. Three lines to add back the day a version lands.

internal/provider/freehire/transport.go
amikai: Live again at `25d9e40` — reopening. Note this one survives dropping the `UserAgent` field: `Base` and the `Clone` contract still need a test either way.

internal/provider/freehire/transport_test.go
amikai: Live again at `25d9e40` — reopening.

internal/provider/freehire/transport_test.go
amikai: Live again at `25d9e40` — reopening. If the `UserAgent` field goes away this collapses to one test rather than a table, which is also fine.

internal/provider/freehire/doc.go
amikai: Live again at `25d9e40` — reopening.

cmd/openings-mcp/main.go
amikai: Live again at `25d9e40` — reopening.

internal/provider/freehire/transport.go
amikai: Withdrawing this one. Keeping the PR to freehire means a shared transport isn't on the table — extracting it drags job104 and eightfold into a change that has nothing to do with them, which is what happened last round. The duplication is real but it isn't this PR's job. Staying resolved.

internal/httpx/transport.go
amikai: Moot — `internal/httpx` no longer exists at `25d9e40`. Resolving.

internal/provider/job104/transport.go
amikai: Moot — `job104/transport.go` is back to its original form with the Referer note intact. Resolving.

internal/provider/freehire/transport.go
amikai: Moot — `UserAgent(version)` is gone at `25d9e40`, and version isn't going in for now. The remaining question about the field itself is back on the `transport.go:14` thread. Resolving.

internal/provider/eightfold/transport.go
amikai: Moot — `eightfold/transport.go` is back to its original form. Resolving.

amikai:
Head is back at `25d9e40`, so the review resets with it. Thread state now:

- **Live again, reopened (6)** — `transport.go:14` (revised suggestion posted there now that version is off the table), `transport.go:24`, `transport_test.go:21`, `transport_test.go:31`, `doc.go:18`, `cmd/openings-mcp/main.go:341`.
- **Withdrawn** — `transport.go:17` (shared transport). Out of scope while this PR stays freehire-only; extracting it is what dragged job104 and eightfold in last round.
- **Closed by decision** — `doc.go:16`. 429 stays unhandled; the host gets `code 429: …` verbatim and can back off on its own.
- **Moot** — the four notes from the `httpx` round; that code no longer exists.

One item still has no diff line to hang on, same as the first review: the `.hurl` fixtures hit freehire.me with hurl's default User-Agent. `make hurl-test` is the one thing here generating repeated live traffic and the one thing not identifying itself, which is backwards given what this PR is for. That fix was in the reverted commit and is worth carrying back — it's freehire-only and touches no other provider.

internal/provider/freehire/transport.go
amikai: Fixed in c01a1fb — field dropped, `RoundTrip` sets `DefaultUserAgent` directly. `Transport` is now the same shape as `eightfold.BrowserTransport`.

internal/provider/freehire/transport.go
amikai: Fixed in c01a1fb — `TestTransportBase` drives `RoundTrip` through a capture RoundTripper and asserts both that the UA reached `Base` and that `Base` saw the clone. `TestTransport` asserts the caller's request is left unmodified.

internal/provider/freehire/transport_test.go
amikai: Fixed in c01a1fb — `http.NewRequestWithContext(t.Context(), ...)` and `t.Cleanup` throughout.

internal/provider/freehire/transport_test.go
amikai: Fixed in c01a1fb — dropping the `UserAgent` field removed the second case outright, so there is no duplicated scaffold left to table.

internal/provider/freehire/doc.go
amikai: Fixed in c01a1fb — `doc.go` removed entirely, so `openapi.yaml` is the only prose left in the package. The README line that pointed at `doc.go` was rewritten so it no longer names a deleted file.

cmd/openings-mcp/main.go
amikai: Fixed in c01a1fb — comment added.

PR #324, opened 2026-09-22
fix: correctness bugs from a full-repo review

amikai:
## Summary
Fixes eight behavior bugs a full-repo review turned up, plus two small error-handling fixes. Each fix is its own commit, and every commit builds and passes vet on its own.

## Key Changes
- **MCP tool errors**: handlers return a plain error instead of `errorResult`, so failed calls no longer carry a zero-valued `structuredContent` (`{"data":null,"total":0}`) that reads as "0 results". Cake and 104 errors now keep the upstream's message.
- **`--http`**: bare `--http` binds `127.0.0.1:8080` instead of every interface. The transport has no auth and the docs already said loopback. Shutdown now waits at most 10s.
- **Workday**: a URL for another site of a roster tenant keeps its own site instead of silently searching the roster's.
- **Eightfold**: for tenants that only have the v2 API (Vale, Bayer, HSBC), `Filters` returns an empty set, and a filtered search asks to retry without filters, instead of a raw 403.
- **ADP MyJobs**: an expired session token is refreshed and the request retried once, instead of failing until the server restarts.
- **ASUS**: descriptions decode every HTML entity (`&#39;`, `&#34;` leaked through before).
- **BambooHR**: detail descriptions for query searches are cached, so paging or re-querying no longer fetches every job's detail again.
- **Lever**: roster lookup ignores case (77 roster sites are mixed-case), which also fixes `cmd/lever --site` for those sites.

## Verification & Testing
New tests: error results carry no structured content; Workday site fold; the Eightfold v2-only tenant path, which had no coverage before; ADP expired-token refresh; ASUS entity decoding; BambooHR request counting; Lever case folding.

`go test -race -count=1 ./...` passes.


amikai:
I found two correctness gaps in this PR. The affected package tests pass, but they do not cover these cases.

1. **[P2] Distinguish the v2-only Eightfold response from other 403s.** `isPCSXDisabled` in `internal/ats/eightfold.go:169-171` checks only the status code. The new `Filters` and `resolveFilters` branches therefore turn *any* PCSX 403 into an empty filter set or advice to retry without filters. A tenant receiving a different authorization or edge 403 would have its real failure hidden. Please match the documented `PCSX is not enabled for this user` response (or otherwise confirm v2-only capability), and test an unrelated 403.

2. **[P2] Keep BambooHR's described dump in sync with its list dump.** `internal/ats/bamboohr.go:138-151` gives the enriched copy a separate write-based TTL. If `Filters` loads the list at t=0 and a queried search fills the described cache at t=9m, the list expires at t=10m while queried searches can still serve the t=0 snapshot until t=19m (with the default 10m TTL). During that interval queried and unqueried searches disagree, and queried searches can miss new jobs. Please expire or invalidate the described entry with the underlying list generation, and cover this expiry boundary.

